PR #25, opened 2019-06-19
Implement engine manager service and add gRPC client

shuo-wu:


engine/service.go
```diff
@@ -179,20 +201,191 @@ func (s *Service) FrontendShutdownCallback(ctx context.Context, req *rpc.EngineR
...
-	if err := e.shutdownFrontendCallback(); err != nil {
+	tID, err := e.shutdownFrontendCallback()
+	if err != nil {
 		return nil, errors.Wrapf(err, "failed to execute stopFrontendCallback for %v", name)
 	}
+
+	s.engines[name] = e
+
+	if tID != 0 {
+		if err := s.availableTIDs.ReleaseRange(tID, tID+1); err != nil {
```
yasker: This should be `ReleaseRange(tID, tID)`. It's all inclusive, see https://github.com/longhorn/longhorn-engine-launcher/blob/processlauncher/util/util.go#L17

engine/service.go
```diff
@@ -2,43 +2,54 @@ package engine
...
+	"os"
 	"strconv"
 	"sync"
 
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 
 	"github.com/longhorn/longhorn-engine-launcher/rpc"
+	emutil "github.com/longhorn/longhorn-engine-launcher/util"
```
yasker: what's wrong with the name `util`?

yasker: I see from a previous code. Normally we call the util in the same package `util`.

PR #26, opened 2019-06-21
engine manager refactor

shuo-wu:


engine/engine_manager.go
```diff
@@ -0,0 +1,313 @@
...
+	for _, el := range em.engineLaunchers {
+		ret.Engines[el.currentEngine.EngineName] = el.RPCResponse()
+	}
+
+	logrus.Infof("Engine Manager has successfully list all engines")
+
+	return ret, nil
+}
+
+func (em *Manager) EngineUpgrade(ctx context.Context, req *rpc.EngineUpgradeRequest) (ret *empty.Empty, err error) {
```
yasker: Separate this function into multiple functions. There are too many locking happens here.

engine/engine.go
```diff
@@ -1,74 +1,78 @@
...
 	"github.com/longhorn/longhorn-engine-launcher/rpc"
+	"github.com/longhorn/longhorn-engine-launcher/util"
 )
 
 type Engine struct {
-	lock *sync.RWMutex
-
-	Name       string
-	VolumeName string
+	EngineName string
```
yasker: Do we need this EngineName? Previously controller has an ID but not name. And I saw the reference of Engine Name below is in fact volume name?

shuo-wu: We need to generate engine process name somewhere. I think it's better to generate it outside of engine manager so that we can use that name to verify the engine creation.

We will use Engine controller to enable the frontend of the corresponding volume. Hence the name of rpc calls are VolumeXXXX.

engine/engine_launcher.go
```diff
@@ -241,35 +229,73 @@ func (el *Launcher) rollbackUpgrade(processLauncher rpc.LonghornProcessLauncherS
...
+		logrus.Errorf("engine launcher %v: failed to deleted backup engine process", el.LauncherName)
+	}
 
-	el.binaryBackup = ""
-	el.backupEngine = nil
+	return
+}
 
-	logrus.Debugf("engine launcher %v: finalize completed", el.LauncherName)
+func (el *Launcher) waitForEngineProcessDeletion(processLauncher rpc.LonghornProcessLauncherServiceServer, processName string) bool {
```
yasker: Feel like we don't need to wait for failure. If it failed, print out an error and continue. Don't need to return false. If a process cannot be killed by SIGKILL we would have a problem, but we cannot do much about it.

PR #27, opened 2019-07-02
Rework logging for process manager

ttpcodes:
This PR makes changes to how logs are handled in `longhorn-engine-launcher` as part of the changes in the `processlauncher` refactor:
- A custom formatter has been created using `logrus` that will use a field value to append the `Process` name as a prefix to the logs of the launcher, making them more readable. Note that if no value is specified for this field, the value will simply default to `longhorn-engine-launcher`.
- `Process` logs are pushed through a special `Writer` interface implementation that will not only log the messages that come through to the `standard output`, but also open a `File` that will write the logs onto the system. The `File` will automatically be cleaned up as the `Process` is deleted.
- Logs are obtainable over `gRPC`, which provides functions that will return a `Stream` that can be parsed to read the logs line-by-line.


engine/engine_manager.go
```diff
@@ -255,6 +255,31 @@ func (em *Manager) EngineUpgrade(ctx context.Context, req *rpc.EngineUpgradeRequ
...
+	em.lock.RLock()
+	defer em.lock.RUnlock()
+
+	el, exists := em.engineLaunchers[req.Name]
+	if !exists {
+		return fmt.Errorf("cannot find engine %v", req.Name)
+	}
+
+	el.lock.RLock()
+	err := el.engineLog(&rpc.LogRequest{
```
yasker: Is it possible to get all the logs happened in this engine launcher instead of just the current one?

ttpcodes:
Removed extra commit that fixed a logging issue that was addressed in #28.

engine/engine_manager.go
yasker: I am going to merge this now. Let's deal with launcher's engine logs later.

PR #29, opened 2019-07-10
Rename process manager with some modifications

shuo-wu:
1. Add new gRPC service `InstanceManagerService` for common functions 
2. Refactor graceful shutdown
3. Check and cleanup frontend before unregistering engine launcher
2. Renaming
3. Adjust log


engine/engine_manager.go
```diff
@@ -107,9 +107,33 @@ func (em *Manager) unregisterEngineLauncher(launcherName string) {
 	if _, err := em.processManager.ProcessGet(nil, &rpc.ProcessGetRequest{
 		Name: processName,
 	}); err != nil && strings.Contains(err.Error(), "cannot find process") {
+		// cannot depend on engine process's callback to cleanup frontend. need to double check here
+		em.lock.RLock()
+		el, exists := em.engineLaunchers[launcherName]
+		em.lock.RUnlock()
```
yasker: I don't think you can unlock here. el is still being used later.

PR #35, opened 2019-07-16
Add streaming methods to gRPC API

ttpcodes:
This PR adds two new methods, `EngineStream` and `ProcessStream`, to the `gRPC` definition for the purpose of sending real-time streaming updates regarding any changes to `Engines` and `Processes` on the `Instance Manager`, needed for the implementation of longhorn/longhorn#631.

Both of these methods are set up to wrap the `channels` used for firing update information on their respective `Managers` such that when an update to an `Engine` or `Process` is detected, the `Manager` will read the updated information, convert it to a `gRPC response`, and then send it to all of the connected `gRPC clients`. Both `streams` are set up to be long-lived and should only close on `error` or if the client does so. These methods are also wrapped using special `structs` defined in the `api package` to avoid having to import the `gRPC definitions` directly.


engine/engine_manager.go
```diff
@@ -318,6 +350,48 @@ func (em *Manager) EngineLog(req *rpc.LogRequest, srv rpc.EngineManagerService_E
...
+		})
+		if err != nil {
+			return err
+		}
+		initialEngineState = append(initialEngineState, e.RPCResponse(response))
+	}
+	em.lock.RUnlock()
+
+	for _, resp := range initialEngineState {
+		if err := srv.Send(resp); err != nil {
```
yasker: It's maybe faster if we stream a list instead of objects one by one. Depends on the protocol, there can be less overhead in transmitting.

engine/engine_manager.go
```diff
@@ -318,6 +350,48 @@ func (em *Manager) EngineLog(req *rpc.LogRequest, srv rpc.EngineManagerService_E
...
+			return err
+		}
+	}
+
+	for resp := range responseChan {
+		if err := srv.Send(resp); err != nil {
```
yasker: Kubernetes retransmit all the states every 30 seconds as you know just in case the delta got wrong sometimes. We may need to do it too.

engine/engine_manager.go
ttpcodes: Would it be faster to stream a `list` of `objects`? I feel like in the case of there being numerous different `objects` updating within a short time-frame, sending multiple `list` responses to the `Client` and having to parse through multiple `lists` of all `objects` versus just the updated `objects` would be slower on the `client` end.

We could collect only the updated `objects` and send that, but then we'd need to specify some time interval to wait for updated `objects` and then send versus just sending off any updates immediately.

Using the `list` only to report the initial `State` from the `Server` doesn't make sense.

Between the current approach and collecting updates within a certain time interval to send off, which do you think makes more sense?

engine/engine_manager.go
ttpcodes: Yeah, this makes sense. Otherwise the client would have to guarantee being able to update the `object` state without errors or it loses the update, which doesn't make sense realistically.

engine/engine_manager.go
yasker: What I was referring to was the extra protocol wrapper needed for each object. I don't have a definitive answer but normally one big package took less time transmitting in the network, though it would depend on the protocol implementation. 

I am OK with the current approach. If we benchmark it later and found some room to improve, we can always do it later.

ttpcodes: I'll leave the current implementation of this as in then, and then revisit it if the benchmarks suggest room for improvement as you say.

engine/engine_manager.go
ttpcodes: Since the logic for getting a live `Watch` on the `objects` and re-polling the status manually every `x` seconds is different, this is something that'll probably be left to the client (`longhorn-manager`) to implement. We can likely have one method responsible for streaming real-time changes and another responsible for checking for the definitive `list` to represent the status every `x` seconds.

ttpcodes:
- Refactored the existing methods to be named `Watch` instead of `Stream`.
- `Watch` methods on the `Server` no longer send the initial state of the `objects` to the `Client`.
- `Client` no longer closes the `GRPC connection` or cancels the `Context` immediately.
- `api` package wrappers now have a `Close` method for cleaning up the `Watch`.

yasker:
What issue does `Client no longer closes the gRPC connection or cancels the Context immediately.` cause?

yasker:
LGTM. We can merge it in once we've tested it end to end with the manager change.

ttpcodes:
The `gRPC Connection` and `Context` issue is due to how `gRPC` handles `Streams`. Rather than blocking once you ask for data from a `Server Stream`, the `Client` receives a `Stream struct`, which contains methods such as `Recv()` that can be used to retrieve and parse data from the `Stream`. 

This is still treated as part of this same `Connection` and `Context` though, meaning that if the `Connection` is closed or the `Context` is cancelled right after retrieving the `Stream`, any attempts to read data out of that `Stream` will raise errors about either the `Context` being canceled or the `Connection` being closed. 

This means that unlike the other methods in the `gRPC Client` (which are referred to by `gRPC` as `Unary` versus `Streaming`), we can't just defer a `cancel()` and `Close()` call immediately after retrieving the `Client` end of the `Stream`, and we have to wrap our own `Close()` call that we can manually call once we're done working with the data in the `Stream`.

ttpcodes:
- Modified `unregister` logic on the `ProcessManager` and `EngineManager` to fire to the `UpdateCh` to indicate `Process` or `Engine` deletion.

ttpcodes:
- Modified some of the `EngineManager lock` calls to avoid potential `deadlock` from sending `Watch` updates (which require acquiring the `locks` of the specific `EngineLauncher` and the `EngineManager`.

PR #36, opened 2019-07-20
Adjust frontend check and cleanup for engine manager

shuo-wu:


engine/engine_launcher.go
```diff
@@ -360,20 +360,20 @@ func (el *Launcher) startFrontend(frontend string) error {
 
 	el.lock.Lock()
 
-	if el.Frontend == frontend && el.Frontend != "" && el.scsiDevice != nil {
-		logrus.Debugf("Engine frontend %v is already up", el.Frontend)
+	if el.Frontend != "" && el.scsiDevice != nil {
```
yasker: Why skip the verification that the frontend is the same?

shuo-wu: The logic is the same. We will do nothing as long as `el.Frontend` is set. And the log message will tell whether the frontend is the same.

yasker: What if the user wants to enable a different frontend but we didn't error out? The end result would not be what the user expected.

PR #40, opened 2019-07-30
Add resource version in instance manager

shuo-wu:


rpc/rpc.proto
```diff
@@ -34,6 +34,7 @@ message EngineSpec {
 message EngineStatus {
 	ProcessStatus process_status = 1;
 	string endpoint = 2;
+	int64 revision_counter = 3;
```
yasker: Don't call it `revision_counter`, just `revision` or `resource_version`.

We're going to get rid of `revision_counter` in the replica.

shuo-wu: Why do we need to get rid of it in the replica? We also need to make sure that the latest replica process status won't be overwritten by expired status.

yasker: I meant in the replica data path. Totally different thing. https://github.com/longhorn/longhorn/issues/508

shuo-wu: Do you mean the variable name `revision_counter` had been used in the replica?

PR #43, opened 2019-08-02
Use assigned UUID rather than self-generated UUID

shuo-wu:


engine/engine_launcher.go
```diff
@@ -73,7 +72,7 @@ type Launcher struct {
 
 func NewEngineLauncher(spec *rpc.EngineSpec) (*Launcher, *Engine) {
 	el := &Launcher{
-		UUID:         uuid.NewV4().String(),
+		UUID:         spec.Uuid,
```
yasker: In fact it should be spelled as `UUID` since it's short for universally unique identifier

shuo-wu: This name is generated by protoc. If we rename `uuid` to `u_u_i_d` in the pb file, we can get `UUID`

yasker:
Should we request UUID if the user wants to operate on the object? To make sure it's operating on the right object.

shuo-wu:
Longhorn engine will call frontend callback and it has no idea about `UUID`.

yasker:
> Longhorn engine will call frontend callback and it has no idea about `UUID`.

It knows. Launcher ID was UUID before.

yasker:
We can do it later though. Not a priority at the moment.

shuo-wu:
> We can do it later though. Not a priority at the moment.

But we need to modify the grpc definition. Hence it cannot be delayed.

PR #78, opened 2020-07-11
Add support for arm64 (draft)

ivang:
Addresses partially longhorn/longhorn#6 by adding support for arm64 (armhf not supported).

This PR should be viewed as part of longhorn/longhorn-engine#514.


yasker:
Close this PR since the change has been included in https://github.com/longhorn/longhorn-instance-manager/pull/83

PR #80, opened 2020-08-28
🐛 Utils: make GetURL IPv6 ready

Frankkkkk:
Hi,
Another small MR to make `GetURL` IPv6 ready for IPv6-only cluster. This function is actually used by the `longhorn-manager` code

Once this MR is merged, I'll be able to make another one in longhorn-manager (https://github.com/Frankkkkk/longhorn-manager/commit/aaec3d48717247e8b28e1e2a51f9aa313f5122ee , still WIP)

Related to https://github.com/longhorn/longhorn-manager/pull/663 and https://github.com/longhorn/longhorn-ui/pull/320

Cheers !

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"net"
 	"os"
 	"os/exec"
+	"strconv"
 	"time"
 
 	"github.com/satori/go.uuid"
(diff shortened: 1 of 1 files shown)


yasker:
Thanks @Frankkkkk . The instance manager is a special piece in Longhorn, which we want to keep it as stable as possible. To use the updated instance manager, remember to change https://github.com/longhorn/longhorn-manager/blob/master/deploy/install/02-components/01-manager.yaml#L30

PR #83, opened 2020-10-14
6 arm support for instance manager

boknowswiki:
https://github.com/longhorn/longhorn/issues/6

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,6 +1,6 @@
 ---
 kind: pipeline
-name: default
+name: default-amd64
 
 platform:
   os: linux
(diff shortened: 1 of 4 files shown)


meldafrawi:
LGTM

.drone.yml
```diff
@@ -14,7 +14,7 @@ steps:
 
 - name: build
   pull: default
-  image: rancher/dapper:1.11.2
+  image: rancher/dapper:v0.5.3
```
joshimoo, line 17: Is this a downgrade or did the versioning scheme change?

boknowswiki: Good catch, let me update it. Thanks.

joshimoo: No problem, looks like the dapper versioning scheme was changed.
@boknowswiki , @meldafrawi We should make sure to update to the the same version for all of our components.

PR #88, opened 2021-06-20
IM Deadlock fixes - part 1

joshimoo:
This contains: 
- 1 data race
there was a race in the process UpdateChannel assignment

- 1 deadlock fix
the above race leads to a deadlock when the binary couldn't be located.

- 1 nil pointer fix:
there was a nil pointer case, while updating a process that is being
deleted, since when initially checked the process was still in the map
but by the time new process has started the old process had been removed

There will be a **part two that fixes another deadlock issue** as well as adds unit tests for all these cases.
Since the builds would potentially deadlock, I decided to split the PR up till the fix for the final deadlock issue is available.

The explanation for the second deadlock is below, this one could be quite common during parallel engine upgrades for example:
// there was a deadlock when the im.monitor is processing an element
// from the updateChannel, it will try to RLock, to evaluate the existing
// processes. This will deadlock, if during that time a process is
// being replaced, since as part of the replacement. The old process
// when being stopped will try to sent the updated process on the update channel
// while the other scope has acquired a WriteLock.
// Since the IM is waiting on the ReadLock acquisition, while inside of the channel receive
// this will be a total deadlock, since the channel receive can never finish therefore all
// additional sents will be blocked.
// https://github.com/longhorn/longhorn/issues/2697

For the unit tests with a description of each dead lock, have a look here:
https://github.com/joshimoo/longhorn-instance-manager/commit/de3bafdb037aa7c025352a50f5debf8e4ae7aaa3

longhorn/longhorn#2697

### DIFF
--- a/pkg/process/process_manager.go
+++ b/pkg/process/process_manager.go
@@ -468,7 +468,7 @@ func (pm *Manager) initProcessReplace(p *Process) (*Process, error) {
 
 	oldProcess, exists := pm.processes[p.Name]
 	if !exists {
-		return nil, status.Errorf(codes.AlreadyExists, "process %v doesn't exists", p.Name)
+		return nil, status.Errorf(codes.NotFound, "existing process %v doesn't exists", p.Name)
 	}
 
 	if err := pm.allocateProcessPorts(p); err != nil {
--- a/pkg/process/process_manager.go
+++ b/pkg/process/process_manager.go
@@ -404,7 +404,8 @@ func (pm *Manager) ProcessReplace(ctx context.Context, req *rpc.ProcessReplaceRe
 		healthChecker: pm.HealthChecker,
 	}
 
-	if err := pm.initProcessReplace(p); err != nil {
+	existingProcess, err := pm.initProcessReplace(p)
+	if err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/process/process_manager.go
```diff
@@ -404,7 +404,8 @@ func (pm *Manager) ProcessReplace(ctx context.Context, req *rpc.ProcessReplaceRe
 		healthChecker: pm.HealthChecker,
 	}
 
-	if err := pm.initProcessReplace(p); err != nil {
+	existingProcess, err := pm.initProcessReplace(p)
```
shuo-wu, line 407: It's better to rename this as `processToBeReplaced` or `oldProcess`

pkg/process/process_manager.go
```diff
@@ -436,12 +437,20 @@ func (pm *Manager) ProcessReplace(ctx context.Context, req *rpc.ProcessReplaceRe
 		return nil, fmt.Errorf("Failed to start replacement process %v", p.Name)
 	}
 	pm.lock.Lock()
-
-	oldProcess := pm.processes[p.Name]
-	oldProcess.StopWithSignal(terminateSignal)
-	//TODO wait for the old process to stop
-	pm.releaseProcessPorts(oldProcess)
-	logrus.Infof("Process Manager: successfully unregistered old process %v", p.Name)
+	if processToReplace, exists := pm.processes[p.Name]; exists {
```
shuo-wu, line 440: this one is `existingProcess`

pkg/process/process_manager.go
```diff
@@ -436,12 +437,20 @@ func (pm *Manager) ProcessReplace(ctx context.Context, req *rpc.ProcessReplaceRe
...
+			pm.releaseProcessPorts(existingProcess)
+			logrus.Infof("Process Manager: successfully unregistered old process %v", p.Name)
+		} else {
+			logrus.Warnf("Process Manager: replace process %v the existing process with UUID %v must have already been unregistered found process with UUID %v",
+				p.Name, existingProcess.UUID, processToReplace.UUID)
+		}
+	} else {
+		logrus.Warnf("Process Manager: replace process %v the existing process with UUID %v no longer exists",
+			p.Name, existingProcess.UUID)
+	}
```
shuo-wu, line 453: If this mismatching happens and there is no cleanup/rollback here, the existing process will become an orphan running process after the unregistering.

pkg/process/process_manager.go
```diff
@@ -468,7 +468,7 @@ func (pm *Manager) initProcessReplace(p *Process) (*Process, error) {
 
 	oldProcess, exists := pm.processes[p.Name]
 	if !exists {
-		return nil, status.Errorf(codes.AlreadyExists, "process %v doesn't exists", p.Name)
+		return nil, status.Errorf(codes.NotFound, "existing process %v doesn't exists", p.Name)
```
shuo-wu, line 471: `old process %v doesn't exists`

shuo-wu (commented):
In general LGTM.

BTW, this deadlock will lead to goroutine hangs but won't block the whole process manager(longhorn/longhorn#2697), right?

innobead (approved):
LGTM, but Shuo has some questions about the naming. Please resolve them if needed.

joshimoo:
See PR here which has the final deadlock fix:
https://github.com/longhorn/longhorn-instance-manager/pull/89

So I am closing this part, since the commits are contained in there, with the suggestions from @shuo-wu

PR #89, opened 2021-06-21
Fixes issues surrounding process replacement

joshimoo:
data race / deadlock fix
there was a race in the process UpdateChannel assignment
the above race leads to a deadlock when the binary couldn't be located.

nil pointer fix:
there was a nil pointer case, while updating a process that is being
deleted, since when initially checked the process was still in the map
but by the time new process has started the old process had been removed

final deadlock which could be triggered through concurrent calls
while trying todo process replacement, explanation for this deadlock is below, 
this one could be quite common during parallel engine upgrades for example:

// there was a deadlock when the im.monitor is processing an element
// from the updateChannel, it will try to RLock, to evaluate the existing
// processes. This will deadlock, if during that time a process is
// being replaced, since as part of the replacement. The old process
// when being stopped will try to sent the updated process on the update channel
// while the other scope has acquired a WriteLock.
// Since the IM is waiting on the ReadLock acquisition, while inside of the channel receive
// this will be a total deadlock, since the channel receive can never finish therefore all
// additional sents will be blocked.

engine upgrade issue longhorn/longhorn#2697
supersedes PR longhorn/longhorn-instance-manager#88

test image: `joshimoo/longhorn-instance-manager:v1_20210621`

### DIFF
--- a/pkg/process/process_test.go
+++ b/pkg/process/process_test.go
@@ -2,6 +2,7 @@ package process
 
 import (
 	"os"
+	"os/exec"
 	"strconv"
 	"sync"
 	"testing"
@@ -20,8 +21,11 @@ import (
 )
 
 const (
-	RetryCount    = 50
-	RetryInterval = 100 * time.Millisecond
+	RetryCount        = 50
+	RetryInterval     = 100 * time.Millisecond
+	TestBinary        = "any"
+	TestBinaryMissing = "fail-missing-binary"
+	TestBinaryReplace = "replacement"
 )
 
 func Test(t *testing.T) { TestingT(t) }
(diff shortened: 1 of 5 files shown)


innobead (approved):

pkg/process/process_test.go
```diff
@@ -147,107 +152,229 @@ func (s *TestSuite) TestProcessDeletion(c *C) {
...
+	wg := &sync.WaitGroup{}
+	pw := &ProcessWatcher{}
+	for i := 0; i < count; i++ {
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
+			go s.pm.ProcessWatch(nil, pw)
+			name := "test_process_deletion_replace-" + strconv.Itoa(i)
+			assertProcessCreation(c, s.pm, name, TestBinary)
+			assertProcessDeletion(c, s.pm, name)
```
shuo-wu, line 250: When this deletion is verified, the old process is already not in the map when the initial check of the process replacement happens. Then does this test still make sense?

joshimoo: Is a required test, since deletion of the process from the map happens async.
The process deletion is triggered, but not removed from the map till the process has actually successfully stopped.

See: https://github.com/joshimoo/longhorn-instance-manager/blob/9781dad48aa70fd86e1a948dc34c4bc7492b53c2/pkg/process/process_manager.go#L233

This is one problem with the way the processes are handled in the IM currently.
Another problem is the port allocations are not bound to a specific process UUID, so it's possible that one process releases/claims anothers ports.

But for v1.1.2 I did not want to touch the process structures, instead I used these tests to verifiy the expected functionality based on the current design, which I have already identified as flawed.

PR #105, opened 2022-02-08
Implement gRPC proxy

c3y1huang:
https://github.com/longhorn/longhorn/issues/2285


shuo-wu (commented):
In the current implementation, This instance manager proxy needs to understand everything of the longhorn-engine gRPC service, which will make the maintenance become more trivial.

Actually, I am thinking if we can use a single gRPC function instead. The request of this function is like:
```
type ProcessProxyRequest struct {
    method string
    args string
}
```
`args` is the json string of the actual request inputs. Then longhorn managers are responsible for serializing and deserializing the actual requests and responses. In other words, only longhorn managers should understand the longhorn-engine gRPC service.

I am not sure if this suggestion is practicable. If you or others have any other concerns, we can have a further discussion.

shuo-wu (approved):
In general LGTM. I didn't walk through each server API implementation.

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,89 @@
...
+		logsDir: logsDir,
+
+		HealthChecker: &GRPCHealthChecker{},
+	}
+	// help to kickstart the broadcaster
+	c, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	if _, err := p.broadcaster.Subscribe(c, p.broadcastConnector); err != nil {
+		return nil, err
+	}
```
shuo-wu, line 42: BTW, what's this for?

pkg/client/proxy.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	return nil
+}
+
+type ProxyClient struct {
+	ServiceURL string
+	ServiceContext
+}
+
+func NewProxyClient(ctx context.Context, ctxCancel context.CancelFunc, address string, port int) (*ProxyClient, error) {
```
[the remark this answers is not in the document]
c3y1huang: The idea is to have a persistent reusable client for each controller ID, so the connection lifetime doesn't go with each API invocation.

pkg/proxy/proxy.go
c3y1huang: Don't need this. Removed.

app/cmd/start.go
```diff
@@ -121,3 +152,13 @@ func start(c *cli.Context) error {
...
 	return <-shutdownCh
 }
+
+func getProxyAddress(listen string) (string, error) {
+	host, port, err := net.SplitHostPort(listen)
+	if err != nil {
+		return "", err
+	}
+
+	intPort, _ := strconv.Atoi(port)
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

c3y1huang:
I don't know why the drone is failing for arm64 and s390x.

pkg/client/proxy.go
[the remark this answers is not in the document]
shuo-wu: Got it. Thank you for your explanation!

pkg/proxy/replica.go
```diff
@@ -35,7 +36,30 @@ func (p *Proxy) ReplicaList(ctx context.Context, req *rpc.ProxyDirect) (resp *rp
 	}
 	ackResponse(resp.ProxyAck, false)
 
-	// TODO
+	c, err := eclient.NewControllerClient(req.ToController)
```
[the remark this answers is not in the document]
shuo-wu: I mean, the proxy server can have a map that caches and reuses all process clients.

app/cmd/start.go
```diff
@@ -30,6 +33,7 @@ func StartCmd() cli.Command {
 			cli.StringFlag{
 				Name:  "listen",
 				Value: "localhost:8500",
+				Usage: "Start a process server with this value. The proxy server will be listening on the next port.",
```
innobead, line 36: So we will keep the process server? why?

app/cmd/start.go
```diff
@@ -108,10 +105,54 @@ func start(c *cli.Context) error {
...
+		return err
+	}
+
+	proxy, err := proxy.NewProxy(logsDir, shutdownCh)
+	if err != nil {
+		return err
 	}
+	hcProxy := health.NewProxyHealthCheckServer(proxy)
+
+	rpcProxyService, proxyServer, err := util.NewServer("tcp://"+proxyAddress, tlsConfig,
```
innobead, line 127: Need to handle it in the below shutdown hook?

pkg/client/proxy.go
```diff
@@ -0,0 +1,91 @@
...
+	getServiceCtx := func(serviceUrl string) (ServiceContext, error) {
+		connection, err := grpc.Dial(serviceUrl, grpc.WithInsecure())
```
innobead, line 46: We should adopt TLS here, because manager will communicate with instance manager.

pkg/client/proxy.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (c *ProxyClient) ServerVersionGet(serviceAddress string) (version *emeta.VersionOutput, err error) {
+	if serviceAddress == "" {
+		return version, errors.Wrapf(ErrParameter, "failed to get server version")
+	}
+	log := logrus.WithFields(logrus.Fields{"serviceURL": c.ServiceURL})
+	log.Debug("Getting server version via proxy")
+
+	return version, nil
```
innobead, line 110: Didn't see any get call to the remote server?

app/cmd/start.go
innobead: My fault. We still need that for engine/replica process :slightly_smiling_face:

app/cmd/start.go
```diff
@@ -108,10 +105,54 @@ func start(c *cli.Context) error {
...
+
+	shutdownCh := make(chan error)
+
+	// Start proxy server
+	proxyAddress, err := getProxyAddress(listen)
+	if err != nil {
+		return err
+	}
+
+	proxy, err := proxy.NewProxy(logsDir, shutdownCh)
```
innobead, line 121: I think the new Proxy server will be only for engine instance manager only which will communicate with its downstream replicas, but here we just start it regardless of whether it's engine or replica instance manager.

I feel we should skip proxy for instance manager replica pod. WDYT?

pkg/imrpc/proxy.proto
```diff
@@ -0,0 +1,344 @@
...
+	rpc BackupRemove(EngineBackupRemoveRequest) returns (ProxyACK);
+
+	rpc ReplicaAdd(EngineReplicaAddRequest) returns (ProxyACK);
+	rpc ReplicaList(ProxyDirect) returns (EngineReplicaListResponse);
+	rpc ReplicaRebuildingStatus(ProxyDirect) returns (EngineReplicaRebuildStatusResponse);
+	rpc ReplicaVerifyRebuild(EngineReplicaVerifyRebuildRequest) returns (ProxyACK);
+	rpc ReplicaRemove(EngineReplicaRemoveRequest) returns (ProxyACK);
+}
+
+message ProxyDirect{
```
innobead, line 46: This is not quite straight forward. How about `ProxyEngine`?

pkg/imrpc/proxy.proto
```diff
@@ -0,0 +1,344 @@
...
+	string to_controller = 1;
```
innobead, line 47: Can we just rename to `address` given the message is `ProxyEngine`?

pkg/imrpc/proxy.proto
```diff
@@ -0,0 +1,344 @@
+syntax="proto3";
+
+package imrpc;
+
+import "google/protobuf/empty.proto";
+import "google/protobuf/timestamp.proto";
+import "github.com/longhorn/longhorn-engine/proto/ptypes/controller.proto";
+import "github.com/longhorn/longhorn-engine/proto/ptypes/syncagent.proto";
+
+service ProxyService {
```
innobead, line 10: Can we just rename to `ProxyEngineService`?

pkg/imrpc/proxy.proto
```diff
@@ -0,0 +1,344 @@
...
+	rpc ReplicaRebuildingStatus(ProxyDirect) returns (EngineReplicaRebuildStatusResponse);
+	rpc ReplicaVerifyRebuild(EngineReplicaVerifyRebuildRequest) returns (ProxyACK);
+	rpc ReplicaRemove(EngineReplicaRemoveRequest) returns (ProxyACK);
+}
+
+message ProxyDirect{
+	string to_controller = 1;
+}
+
+message ProxyACK {
```
innobead, line 50: Given other comments. suggest renaming to `ProxyEngineResponse`

pkg/imrpc/proxy.proto
```diff
@@ -0,0 +1,344 @@
...
+	bool complete = 1;
+}
+
+message EngineServerVersionResponse {
```
innobead, line 54: Suggest to `EngineServerVersionProxyResponse`. The same change suggestion for other messages.

innobead (changes requested):
In general, LGTM. Some comments and suggestions needed to reply.

pkg/imrpc/proxy.proto
c3y1huang: Updated.

pkg/imrpc/proxy.proto
c3y1huang: Updated.

pkg/imrpc/proxy.proto
c3y1huang: Updated.

pkg/imrpc/proxy.proto
c3y1huang: Updated.

pkg/imrpc/proxy.proto
c3y1huang: Updated.

app/cmd/start.go
c3y1huang: I think we can enhance this later. Added comment.

innobead: Sure, please create a ticket for this later.

pkg/imrpc/proxy.proto
innobead: Sorry, it should be `ProxyEngineRequest` instead.

innobead:
> In the current implementation, This instance manager proxy needs to understand everything of the longhorn-engine gRPC service, which will make the maintenance become more trivial.
> 
> Actually, I am thinking if we can use a single gRPC function instead. The request of this function is like:
> 
> ```
> type ProcessProxyRequest struct {
>     method string
>     args string
> }
> ```
> 
> `args` is the json string of the actual request inputs. Then longhorn managers are responsible for serializing and deserializing the actual requests and responses. In other words, only longhorn managers should understand the longhorn-engine gRPC service.
> 
> I am not sure if this suggestion is practicable. If you or others have any other concerns, we can have a further discussion.

Good point, this concept is really like proxy implementation. From the development point of view, it will be dynamic but also probably error-prone if any requests from longhorn manager are incorrect like typo. The static typing will be much safe, but tedious and need to have the interface knowledge as well.

I would say let's keep it as it first, because right now this design should not bring any major concerns at the moment, but we can keep improving the proxy part in coming patches.

app/cmd/start.go
c3y1huang: https://github.com/longhorn/longhorn/issues/3968

pkg/imrpc/proxy.proto
c3y1huang: Updated.

Dockerfile.dapper
```diff
@@ -55,6 +57,16 @@ RUN cd /usr/src && \
     make && \
     make install
 
+# Install libqcow to resolve error:
+#   vendor/github.com/longhorn/longhorn-engine/pkg/qcow/libqcow.go:6:11: fatal error: libqcow.h: No such file or directory
```
innobead, line 61: Can we just use installing libqcow-devel package?

I am also confused why the previous builds work?

pkg/meta/version.go
```diff
@@ -4,6 +4,9 @@ const (
 	// InstanceManagerAPIVersion used to communicate with the user e.g. longhorn-manager
 	InstanceManagerAPIVersion    = 1
 	InstanceManagerAPIMinVersion = 1
+
```
innobead, line 7: Add a comment to describe the purpose?

c3y1huang: Added.

Dockerfile.dapper
c3y1huang: I am not sure. I'm using the same install as the engine
https://github.com/longhorn/longhorn-engine/blob/045265cd71d1d3c01ead215da49c1a0c1b40ebaf/Dockerfile.dapper#L51-L58

Maybe we can have another issue to deal with this.

innobead: Yeah, let's tackle another issue, but it seems right now longhorn-instance-manager depends on longhorn-engine proto.

go.mod
```diff
@@ -6,13 +6,13 @@ require (
 	github.com/RoaringBitmap/roaring v0.4.18
 	github.com/golang/protobuf v1.3.3-0.20190920234318-1680a479a2cf
 	github.com/google/uuid v1.3.0
-	github.com/pkg/errors v0.8.1
-	github.com/sirupsen/logrus v1.4.1
+	github.com/longhorn/backupstore v0.0.0-20211103081249-ebc3c90895ba
+	github.com/longhorn/longhorn-engine v1.2.4
```
innobead, line 10: This is tricky, because generate_grpc.sh using the master of longhorn-engine, but here use v1.2.4. 

Would like to see Shuo's feedback.

pkg/proxy/backup.go
```diff
@@ -0,0 +1,444 @@
...
+
+	resp = &rpc.EngineSnapshotBackupProxyResponse{
+		ProxyEngineResponse: &rpc.ProxyEngineResponse{},
+		BackupId:            "",
+		Replica:             "",
+		IsIncremental:       false,
+	}
+	ackResponse(resp.ProxyEngineResponse, false)
+
+	newCtx, cancel := context.WithCancel(context.Background())
```
innobead, line 39: Why not just use `ctx` instead of creating another `newCtx`? the same question to other methods.

pkg/proxy/backup.go
```diff
@@ -0,0 +1,444 @@
...
+	defer cancel()
+
+	for _, env := range req.Envs {
+		part := strings.SplitN(env, "=", 2)
+		if err := os.Setenv(part[0], part[1]); err != nil {
+			return resp, err
+		}
+	}
```
innobead, line 47: This way will impact the environment variables of instance manager process? This will be different from the previous implementation: engineapi will start a process to run engine binary, so it can be launched with environment variables. (note: we only have one backup target now, so probably it will not have any impact, but still uncertain)

Probably you need to have a new GetBackupCredential which is getting info from the passed envs instead of os environment variables. The same for other places.

innobead (changes requested):

pkg/client/proxy.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type ProxyClient struct {
+	ServiceURL string
+	ServiceContext
+}
+
+func NewProxyClient(ctx context.Context, ctxCancel context.CancelFunc, address string, port int) (*ProxyClient, error) {
+	getServiceCtx := func(serviceUrl string) (ServiceContext, error) {
+		connection, err := grpc.Dial(serviceUrl, grpc.WithInsecure())
```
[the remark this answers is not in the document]
innobead: cc @joshimoo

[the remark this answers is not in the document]
c3y1huang: https://github.com/longhorn/longhorn/issues/3975

pkg/client/proxy_backup.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+	recv, err := c.service.SnapshotBackup(c.ctx, req)
+	if err != nil {
+		return "", "", errors.Wrapf(err, "failed to backup snapshot %v to %v via proxy %v to %v", snapshotName, backupName, c.ServiceURL, serviceAddress)
+	}
+
+	return recv.BackupId, recv.Replica, nil
+}
+
+func (c *ProxyClient) SnapshotBackupStatus(serviceAddress, backupName, replicaAddress string) (status *SnapshotBackupStatus, err error) {
```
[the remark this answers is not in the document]
c3y1huang: This ensures the returning object is able to use the fallback clients.

Dockerfile.dapper
c3y1huang: https://github.com/longhorn/longhorn/issues/3976

Dockerfile.dapper
```diff
@@ -55,6 +57,16 @@ RUN cd /usr/src && \
...
+RUN wget -O - https://s3-us-west-1.amazonaws.com/rancher-longhorn/libqcow-alpha-20181117.tar.gz | tar xvzf - -C /usr/src
```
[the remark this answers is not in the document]
c3y1huang: https://github.com/longhorn/longhorn/issues/3976

pkg/proxy/backup.go
c3y1huang: https://github.com/longhorn/longhorn/issues/3977

Note, `SyncAgentServer` will still update the os env per request, for example, [BackupCreate](https://github.com/longhorn/longhorn-engine/blob/master/pkg/sync/rpc/server.go#L676-L701).

go.mod
c3y1huang: Updated to master(v1.3.0-preview1).

pkg/proxy/backup.go
c3y1huang: Fixed.

pkg/client/proxy_backup.go
```diff
@@ -0,0 +1,342 @@
...
+	for _, re := range e.ReplicaErrors {
+		errs = append(errs, re.Error())
+	}
+
+	if errs == nil {
+		return "Unknown"
+	}
+	if len(errs) == 1 {
+		return errs[0]
+	}
```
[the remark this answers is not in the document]
c3y1huang: Removed.

pkg/proxy/backup.go
```diff
@@ -0,0 +1,387 @@
...
+	resp = &rpc.ProxyEngineResponse{}
+	ackResponse(resp, false)
+
+	for _, env := range req.Envs {
+		part := strings.SplitN(env, "=", 2)
+		os.Setenv(part[0], part[1])
+	}
+
+	if req.VolumeName == "" {
+		if err := backupstore.DeleteDeltaBlockBackup(req.DestUrl); err != nil {
```
innobead, line 421: Need to use `destURL = util.UnescapeURL(destURL)` the same as backupstore command implementation.

c3y1huang: Not quite sure why. This is not inputted from bash.
https://github.com/longhorn/longhorn-engine/blob/master/pkg/util/util.go#L248

innobead: yea, the point is:

As per our discussion, to avoid any missing logic when copying the code from engine/backup commands to proxy functions, I would suggest moving the shared part from commands to shared packages which can then be leveraged by commands and proxy, so it will definitely ensure nothing missing in proxy implementation. 

This is quite important, because all volume/snapshot/backup operations will be through the proxy later, so we should ensure the result is the same as running commands. 

Please create a ticket to improve it in the next sprint. Thanks.

c3y1huang: https://github.com/longhorn/longhorn/issues/3981

innobead (approved):
LGTM, but we have some following issues needed to tackle in the next sprint and need to have them in 1.3.0. Thanks, @c3y1huang . 

See @shuo-wu has any concerns before merging this.

pkg/client/proxy.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (c *ProxyClient) Ping() (err error) {
+	_, err = c.service.Ping(c.ctx, &empty.Empty{})
+	if err != nil {
+		return errors.Wrapf(err, "failed to ping %v proxy server", c.ServiceURL)
+	}
+	return nil
+}
+
+func (c *ProxyClient) ServerVersionGet(serviceAddress string) (version *emeta.VersionOutput, err error) {
```
shuo-wu, line 83: As mentioned in the manager PR, need to add function `ClientVersionGet` as well.

pkg/client/proxy_backup.go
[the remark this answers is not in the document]
shuo-wu: Considering the backward compatibility, I didn't directly use the proto-generated struct outside of the gRPC client/server since it may change version by version.

pkg/client/proxy_backup.go
```diff
@@ -71,15 +75,64 @@ func (c *ProxyClient) SnapshotBackupStatus(serviceAddress, backupName, replicaAd
 	return status, nil
 }
 
-func (c *ProxyClient) BackupRestore(serviceAddress, url, target, volumeName string, envs []string) (taskError []byte, err error) {
+type TaskError struct {
```
shuo-wu, line 79: NIT: Put these in a separate go file.

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,70 @@
...
+			done = true
+			break
+		}
+		if done {
+			break
+		}
+	}
+}
+
+func ackResponse(proxyEngineResponse *rpc.ProxyEngineResponse, complete bool) {
```
shuo-wu, line 47: Just curious, what is this for?

pkg/proxy/replica.go
```diff
@@ -35,7 +36,30 @@ func (p *Proxy) ReplicaList(ctx context.Context, req *rpc.ProxyEngineRequest) (r
 	}
 	ackResponse(resp.ProxyEngineResponse, false)
 
-	// TODO
+	c, err := eclient.NewControllerClient(req.Address)
```
shuo-wu, line 53: As I mentioned before, can we store/cache the client in this proxy server? The idea should be similar to `ProxyHandler` in the manager PR. If YES, you can create a ticket & add a TODO.

go.mod
shuo-wu: IIRC, we plan to switch to the version-format image tag rather than the data-format tag for instance manager. Then in the future, we can have one-one matching. For now I think we need to use v1.3.0 longhorn-engine repo.

shuo-wu (changes requested):
Same. I didn't walk through each API implementation since I would assume most of them are similar to the previous implementation in the longhorn-engine or longhorn-manager.
If there are any changes you are not confident with, please notify me.

pkg/client/proxy.go
c3y1huang: Added.

pkg/client/proxy_backup.go
c3y1huang: Moved.

pkg/proxy/proxy.go
c3y1huang: I forgot why I had those. Removed.

pkg/proxy/replica.go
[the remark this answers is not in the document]
c3y1huang: https://github.com/longhorn/longhorn/issues/3984

pkg/proxy/replica.go
c3y1huang: https://github.com/longhorn/longhorn/issues/3984

pkg/client/proxy.go
```diff
@@ -0,0 +1,146 @@
...
+		CLIAPIMinVersion:        int(serverVersion.CliAPIMinVersion),
+		ControllerAPIVersion:    int(serverVersion.ControllerAPIVersion),
+		ControllerAPIMinVersion: int(serverVersion.ControllerAPIMinVersion),
+		DataFormatVersion:       int(serverVersion.DataFormatVersion),
+		DataFormatMinVersion:    int(serverVersion.DataFormatMinVersion),
+	}
+	return version, nil
+}
+
+func (c *ProxyClient) ClientVersionGet(serviceAddress string) (version *emeta.VersionOutput, err error) {
```
shuo-wu, line 117: Why is this cli version get implementation the same as the server version?
This should not involve in a gRPC call. Instead, this client should directly read the version number from pkg `meta`

shuo-wu (changes requested):

pkg/client/proxy.go
c3y1huang: Fixed.

pkg/proxy/replica.go
```diff
@@ -30,7 +31,30 @@ func (p *Proxy) ReplicaList(ctx context.Context, req *rpc.ProxyEngineRequest) (r
 		ReplicaList: &eptypes.ReplicaListReply{},
 	}
 
-	// TODO
+	c, err := eclient.NewControllerClient(req.Address)
+	if err != nil {
+		return resp, err
```
shuo-wu, line 50: NIT: Explicitly returning `nil` rather than `resp` if there is an error. Similar for other implementations

pkg/imrpc/proxy.proto
```diff
@@ -0,0 +1,309 @@
...
+	rpc ReplicaVerifyRebuild(EngineReplicaVerifyRebuildRequest) returns (google.protobuf.Empty);
+	rpc ReplicaRemove(EngineReplicaRemoveRequest) returns (google.protobuf.Empty);
+}
+
+message ProxyEngineRequest{
+	string address = 1;
+}
+
+message EngineVersionProxyResponse {
+	ptypes.VersionOutput version = 2;
```
shuo-wu, line 51: hmmmm... Not sure why the number starts from 2. Similar below

pkg/proxy/replica.go
```diff
@@ -30,7 +31,30 @@ func (p *Proxy) ReplicaList(ctx context.Context, req *rpc.ProxyEngineRequest) (r
 		ReplicaList: &eptypes.ReplicaListReply{},
 	}
```
shuo-wu, line 46: NIT: there is no need to init `resp` at the beginning.

shuo-wu (approved):
In general LGTM

pkg/proxy/replica.go
c3y1huang: I don't think gRPC server can return a nil message.

pkg/imrpc/proxy.proto
c3y1huang: Fixed.

pkg/proxy/replica.go
c3y1huang: Fixed.

pkg/proxy/replica.go
c3y1huang: Updated.

app/cmd/start.go
shuo-wu: After re-checking this part. It seems we do not really need the proxying for IM-R. And the IM-E proxy server should be able to directly use the replica gRPC clients instead.
Since it's the IM-R proxy server rather than the longhorn manager responsible for talking with the replica processes (for backup-related tasks).

pkg/proxy/replica.go
```diff
@@ -96,7 +96,15 @@ func (p *Proxy) ReplicaRemove(ctx context.Context, req *rpc.EngineReplicaRemoveR
...
 
-	// TODO
+	c, err := eclient.NewControllerClient(req.ProxyEngineRequest.Address)
+	if err != nil {
+		return nil, err
+	}
+	defer c.Close()
+
+	if err = c.ReplicaDelete(req.ReplicaAddress); err != nil {
+		return resp, nil
```
shuo-wu, line 132:
```
return nil, err
```

pkg/proxy/volume.go
```diff
@@ -18,7 +19,28 @@ func (p *Proxy) VolumeGet(ctx context.Context, req *rpc.ProxyEngineRequest) (res
 		Volume: &eptypes.Volume{},
 	}
```
shuo-wu, line 20: NIT: same here. don't need to init the response at the beginning.

pkg/proxy/replica.go
```diff
@@ -78,7 +78,25 @@ func (p *Proxy) ReplicaRebuildingStatus(ctx context.Context, req *rpc.ProxyEngin
 		Status: make(map[string]*eptypes.ReplicaRebuildStatusResponse),
 	}
```
shuo-wu, line 79: NIT: same here. don't need to init the response at the beginning.

pkg/proxy/snapshot.go
```diff
@@ -18,7 +19,18 @@ func (p *Proxy) VolumeSnapshot(ctx context.Context, req *rpc.EngineVolumeSnapsho
 		Snapshot: &eptypes.VolumeSnapshotReply{},
 	}
```
shuo-wu, line 21: NIT: same here. don't need to init the response at the beginning.

pkg/proxy/snapshot.go
```diff
@@ -43,7 +44,30 @@ func (p *Proxy) SnapshotList(ctx context.Context, req *rpc.ProxyEngineRequest) (
 		Disks: map[string]*rpc.EngineSnapshotDiskInfo{},
 	}
```
shuo-wu, line 45: NIT: same here. don't need to init the response at the beginning.

pkg/proxy/backup.go
```diff
@@ -18,7 +25,40 @@ func (p *Proxy) SnapshotBackup(ctx context.Context, req *rpc.EngineSnapshotBacku
 		IsIncremental: false,
 	}
```
shuo-wu, line 36: NIT: same here. don't need to init the response at the beginning.

pkg/proxy/backup.go
```diff
@@ -18,7 +25,40 @@ func (p *Proxy) SnapshotBackup(ctx context.Context, req *rpc.EngineSnapshotBacku
...
+	for k, v := range req.Labels {
+		labels = append(labels, fmt.Sprintf("%s=%s", k, v))
+	}
+
+	task, err := esync.NewTask(ctx, req.ProxyEngineRequest.Address)
+	if err != nil {
+		return nil, err
+	}
+
+	recv, err := task.CreateBackup(
```
shuo-wu, line 60: Should check this error first

pkg/proxy/snapshot.go
```diff
@@ -129,7 +129,18 @@ func (p *Proxy) SnapshotRemove(ctx context.Context, req *rpc.EngineSnapshotRemov
...
-	// TODO
+	task, err := esync.NewTask(ctx, req.ProxyEngineRequest.Address)
+	if err != nil {
+		return nil, err
+	}
+
+	var lastErr error
+	for _, name := range req.Names {
+		if err := task.DeleteSnapshot(name); err != nil {
+			lastErr = err
```
shuo-wu, line 207: NIT: I think we can aggregate the errors.

pkg/proxy/snapshot.go
```diff
@@ -135,7 +135,24 @@ func (p *Proxy) SnapshotPurgeStatus(ctx context.Context, req *rpc.ProxyEngineReq
 		Status: map[string]*eptypes.SnapshotPurgeStatusResponse{},
 	}
```
shuo-wu, line 171: NIT: same here. don't need to init the response at the beginning.

pkg/proxy/snapshot.go
```diff
@@ -104,7 +104,27 @@ func (p *Proxy) SnapshotCloneStatus(ctx context.Context, req *rpc.ProxyEngineReq
 		Status: map[string]*eptypes.SnapshotCloneStatusResponse{},
 	}
```
shuo-wu, line 105: NIT: same here. don't need to init the response at the beginning.

pkg/proxy/backup.go
```diff
@@ -69,7 +74,72 @@ func (p *Proxy) SnapshotBackupStatus(ctx context.Context, req *rpc.EngineSnapsho
 
 	resp = &rpc.EngineSnapshotBackupStatusProxyResponse{}
```
shuo-wu, line 80: NIT: same here. don't need to init the response at the beginning.

pkg/proxy/backup.go
```diff
@@ -69,7 +74,72 @@ func (p *Proxy) SnapshotBackupStatus(ctx context.Context, req *rpc.EngineSnapsho
...
+
+	for _, r := range replicas.ReplicaList.Replicas {
+		if r.Address.Address != replicaAddress {
+			continue
+		}
+		mode := eptypes.GRPCReplicaModeToReplicaMode(r.Mode)
+		if mode != etypes.RW {
+			return nil, errors.Errorf("failed to get % v backup status on unknown replica %s", req.BackupName, replicaAddress)
+		}
+	}
```
shuo-wu, line 129: Looks redundant.

pkg/proxy/backup.go
```diff
@@ -69,7 +74,72 @@ func (p *Proxy) SnapshotBackupStatus(ctx context.Context, req *rpc.EngineSnapsho
...
+	cReplica, err := rclient.NewReplicaClient(replicaAddress)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to create replica client with %v", replicaAddress)
+	}
+	defer cReplica.Close()
+
+	status, err := esync.FetchBackupStatus(cReplica, req.BackupName, replicaAddress)
+	if err != nil {
+		return nil, err
+	}
```
shuo-wu, line 140: We already got the backup status above (in the 1st for loop). Should we repeat it here?

pkg/proxy/backup.go
```diff
@@ -150,7 +151,34 @@ func (p *Proxy) BackupRestore(ctx context.Context, req *rpc.EngineBackupRestoreR
 
 	resp = &rpc.EngineBackupRestoreProxyResponse{}
```
shuo-wu, line 156: NIT: same here. don't need to init the response at the beginning.

pkg/proxy/backup.go
```diff
@@ -150,7 +151,34 @@ func (p *Proxy) BackupRestore(ctx context.Context, req *rpc.EngineBackupRestoreR
...
+		return nil, err
+	}
+
+	err = task.RestoreBackup(req.Url, credential)
+	if err != nil {
+		errInfo, jsonErr := json.Marshal(err)
+		if jsonErr != nil {
+			log.Debugf("Cannot marshal err [%v] to json: %v", err, jsonErr)
+		}
+		// If the error is not `TaskError`, the marshaled result is an empty json string.
```
shuo-wu, line 179: NIT: I am thinking we need to error this error

pkg/proxy/backup.go
```diff
@@ -191,7 +191,28 @@ func (p *Proxy) BackupRestoreStatus(ctx context.Context, req *rpc.ProxyEngineReq
 		Status: map[string]*rpc.EngineBackupRestoreStatus{},
 	}
```
shuo-wu, line 196: NIT: same here. don't need to init the response at the beginning.

pkg/proxy/backup.go
```diff
@@ -251,11 +253,69 @@ func (p *Proxy) BackupVolumeList(ctx context.Context, req *rpc.EngineBackupVolum
 		Volumes: map[string]*rpc.EngineBackupVolumeInfo{},
 	}
```
shuo-wu, line 290: NIT: same here. don't need to init the response at the beginning.

pkg/proxy/backup.go
```diff
@@ -251,11 +253,69 @@ func (p *Proxy) BackupVolumeList(ctx context.Context, req *rpc.EngineBackupVolum
...
 
-	// TODO
+	for _, env := range req.Envs {
+		part := strings.SplitN(env, "=", 2)
```
shuo-wu, line 293: Check length

pkg/proxy/backup.go
```diff
@@ -334,7 +335,23 @@ func (p *Proxy) BackupRemove(ctx context.Context, req *rpc.EngineBackupRemoveReq
 	}
 	log.Debug("Removing backups")
 
-	// TODO
+	for _, env := range req.Envs {
+		part := strings.SplitN(env, "=", 2)
```
shuo-wu, line 387: Check length

pkg/proxy/backup.go
```diff
@@ -241,7 +241,29 @@ func (p *Proxy) BackupVolumeGet(ctx context.Context, req *rpc.EngineBackupVolume
 		Volume: &rpc.EngineBackupVolumeInfo{},
 	}
 
-	// TODO
+	for _, env := range req.Envs {
+		part := strings.SplitN(env, "=", 2)
```
shuo-wu, line 258: Check length

pkg/proxy/backup.go
```diff
@@ -241,7 +241,29 @@ func (p *Proxy) BackupVolumeGet(ctx context.Context, req *rpc.EngineBackupVolume
 		Volume: &rpc.EngineBackupVolumeInfo{},
 	}
```
shuo-wu, line 255: NIT: same here. don't need to init the response at the beginning.

pkg/proxy/backup.go
```diff
@@ -228,7 +228,19 @@ func (p *Proxy) BackupGet(ctx context.Context, req *rpc.EngineBackupGetRequest)
 		Backup: &rpc.EngineBackupInfo{},
 	}
 
-	// TODO
+	for _, env := range req.Envs {
+		part := strings.SplitN(env, "=", 2)
```
shuo-wu, line 233: Check length

pkg/proxy/backup.go
```diff
@@ -228,7 +228,19 @@ func (p *Proxy) BackupGet(ctx context.Context, req *rpc.EngineBackupGetRequest)
 		Backup: &rpc.EngineBackupInfo{},
 	}
```
shuo-wu, line 230: NIT: same here. don't need to init the response at the beginning.

pkg/proxy/backup.go
```diff
@@ -357,7 +358,20 @@ func (p *Proxy) BackupConfigMetaGet(ctx context.Context, req *rpc.EngineBackupCo
 
 	resp = &rpc.EngineBackupConfigMetaGetProxyResponse{}
```
shuo-wu, line 359: NIT: same here. don't need to init the response at the beginning.

pkg/proxy/backup.go
```diff
@@ -357,7 +358,20 @@ func (p *Proxy) BackupConfigMetaGet(ctx context.Context, req *rpc.EngineBackupCo
...
 
-	// TODO
+	for _, env := range req.Envs {
+		part := strings.SplitN(env, "=", 2)
```
shuo-wu, line 362: Check length

pkg/proxy/backup.go
```diff
@@ -0,0 +1,405 @@
...
+}
+
+func (p *Proxy) BackupRestore(ctx context.Context, req *rpc.EngineBackupRestoreRequest) (resp *rpc.EngineBackupRestoreProxyResponse, err error) {
+	log := logrus.WithFields(logrus.Fields{"serviceURL": req.ProxyEngineRequest.Address})
+	log.Debugf("Restoring backup %v to %v", req.Url, req.VolumeName)
+
+	resp = &rpc.EngineBackupRestoreProxyResponse{}
+
+	for _, env := range req.Envs {
+		part := strings.SplitN(env, "=", 2)
```
shuo-wu, line 159: check length

shuo-wu (changes requested):
You can ignore the NIT comments first

pkg/proxy/replica.go
c3y1huang: Fixed.

pkg/proxy/volume.go
c3y1huang: Updated

pkg/proxy/replica.go
c3y1huang: Updated.

pkg/proxy/snapshot.go
c3y1huang: Updated.

pkg/proxy/snapshot.go
c3y1huang: Updated.

pkg/proxy/backup.go
c3y1huang: Updated.

pkg/proxy/backup.go
c3y1huang: Fixed.

pkg/proxy/snapshot.go
c3y1huang: Updated.

pkg/proxy/snapshot.go
c3y1huang: Updated.

pkg/proxy/backup.go
c3y1huang: Updated.

c3y1huang: Updated.

pkg/proxy/backup.go
c3y1huang: Removed.

pkg/proxy/backup.go
c3y1huang: Removed.

pkg/proxy/backup.go
c3y1huang: Updated.

pkg/proxy/backup.go
c3y1huang: Updated.

pkg/proxy/backup.go
c3y1huang: Updated.

pkg/proxy/backup.go
c3y1huang: Updated.

pkg/proxy/backup.go
c3y1huang: Updated.

pkg/proxy/backup.go
c3y1huang: Updated.

pkg/proxy/backup.go
c3y1huang: Fixed.

pkg/proxy/backup.go
c3y1huang: Fixed.

pkg/proxy/backup.go
c3y1huang: Fixed.

pkg/proxy/backup.go
c3y1huang: Fixed.

pkg/proxy/backup.go
c3y1huang: Fixed.

pkg/proxy/backup.go
```diff
@@ -0,0 +1,407 @@
...
+)
+
+func (p *Proxy) SnapshotBackup(ctx context.Context, req *rpc.EngineSnapshotBackupRequest) (resp *rpc.EngineSnapshotBackupProxyResponse, err error) {
+	log := logrus.WithFields(logrus.Fields{"serviceURL": req.ProxyEngineRequest.Address})
+	log.Debugf("Backing up snapshots %v to backup %v", req.SnapshotName, req.BackupName)
+
+	for _, env := range req.Envs {
+		part := strings.SplitN(env, "=", 2)
+		if len(part) < 2 {
+			continue
```
shuo-wu, line 35: NIT: it's better to print a log here.

pkg/proxy/backup.go
```diff
@@ -0,0 +1,407 @@
...
+	}
+	defer c.Close()
+
+	replicas, err := p.ReplicaList(ctx, req.ProxyEngineRequest)
+	if err != nil {
+		return nil, err
+	}
+
+	status := &esync.BackupStatusInfo{}
+	replicaAddress := req.ReplicaAddress
```
shuo-wu, line 94: NIT: 
```
var  status *esync.BackupStatusInfo

......

if status == nil {
	return nil, errors.Errorf("failed to find a replica with backup %s", req.BackupName)
}
```

shuo-wu (approved):
LGTM

pkg/proxy/backup.go
innobead: @c3y1huang  you can follow up with this later. Thanks.

pkg/proxy/backup.go
innobead: env part will be refactored, so @c3y1huang will tackle this later.

PR #106, opened 2022-02-18
Add package name to the protocol buffer definition

jenting:
Add package name to the protocol buffer definition file to avoid the namespace conflict.
Use the unique package name `longhorn.instancemanager.pkg.rpc`.

[Longhorn 3343](https://github.com/longhorn/longhorn/issues/3343)

Signed-off-by: JenTing Hsiao <jenting.hsiao@suse.com>


pkg/rpc/rpc.proto
```diff
@@ -1,7 +1,11 @@
 syntax="proto3";
 
+package longhorn.io.longhorn.instance.manager.pkg.rpc;
```
shuo-wu, line 3: Same as the backing image manager PR.
I prefer `longhorn.instancemanager.pkg.rpc`

jenting: Sounds good.

jenting:
Close this one PR since adding the package name to the instance-manager would break the backward compatibility.
For example, the new longhorn-manager can't talk to the old instance-manager because the path changed. Here is the error message shown when longhorn-manager try to connect to the old instance-manager.
```
error receiving next item in engine watch: rpc error: code = Unimplemented desc = unknown service longhorn.instancemanager.pkg.rpc.ProcessManagerService
```

PR #108, opened 2022-04-11
Rename rpc.proto to imrpc.proto

derekbit:
To avoid the "protobuf namespace conflicts" error, renaming the protocol buffer file.

Longhorn 3796


innobead (approved):
LGTM

PR #109, opened 2022-04-11
MTLS authentication implementation

joshimoo:
Tracking issue longhorn/longhorn#3839

### DIFF
--- a/pkg/util/grpcutil.go
+++ b/pkg/util/grpcutil.go
@@ -43,7 +43,7 @@ func Connect(endpoint string, tlsConfig *tls.Config, dialOptions ...grpc.DialOpt
 	// when using Unix domain sockets. It ensures that gRPC detects a dead connection
 	// in a timely manner.
 	// Code lifted from https://github.com/kubernetes-csi/csi-test/commit/6b8830bf5959a1c51c6e98fe514b22818b51eeeb
-	dialOptions = append(dialOptions, grpc.WithKeepaliveParams(keepalive.ClientParameters{PermitWithoutStream: true}))
+	dialOptions = append(dialOptions, grpc.WithKeepaliveParams(keepalive.ClientParameters{Time: 30 * time.Second, PermitWithoutStream: true}))
 
 	return grpc.Dial(address, dialOptions...)
 }
--- a/pkg/util/grpcutil.go
+++ b/pkg/util/grpcutil.go
@@ -1,0 +1,221 @@
+package util
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"os"
+	"strings"
+	"time"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/keepalive"
+)
+
+func unixDialer(ctx context.Context, addr string) (net.Conn, error) {
+	dialer := net.Dialer{}
+	return dialer.DialContext(ctx, "unix", addr)
+}
+
+//Connect is a helper function to initiate a grpc client connection to server running at endpoint using tlsConfig
+func Connect(endpoint string, tlsConfig *tls.Config, dialOptions ...grpc.DialOption) (*grpc.ClientConn, error) {
+	proto, address, err := parseEndpoint(endpoint)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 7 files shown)


joshimoo:
Rebased and updated with the grpc proto conflict fix.
Once this is merge I will update https://github.com/longhorn/longhorn-manager/pull/1292
To point to the new im image and replace the dev -> local ref of the im dep.

pkg/util/grpcutil.go
```diff
@@ -0,0 +1,221 @@
...
+		RootCAs:       certPool,
+	}
+	if peerCert != nil {
+		tlsConfig.Certificates = append(tlsConfig.Certificates, *peerCert)
+	}
+	return tlsConfig
+}
+
+func loadCertificate(caFile, certFile, keyFile string) (certPool *x509.CertPool, peerCert *tls.Certificate, err error) {
+	if certFile != "" || keyFile != "" {
```
innobead, line 190: This should be `if certFile != "" && keyFile != ""`? otherwise it's unable to create the server/client key pair.

innobead (commented):
LGTM, except one comment.

pkg/util/grpcutil.go
```diff
@@ -0,0 +1,221 @@
...
+	certificate, err := tls.X509KeyPair(cert, key)
+	if err != nil {
+		return nil, err
+	}
+
+	return clientConfig(certPool, &certificate, peerName), nil
+}
+
+// LoadClientTLS prepares the TLS configuration that can be used by a client while connecting to a server.
+// peerName must be provided when expecting the server to offer a certificate with that CommonName. caFile, certFile, and keyFile are all optional.
```
innobead, line 166: I believe at least need to provide `caFile`, otherwise how to verify the server connection? 

`certFile` and `keyFile` are needed only when we would like to have mTLS to let sever/client identify each other. I believe we will introduce below cases:
- longhorn-manager -> longhorn-instance-manager/engine via TLS connect not mutual
- longhorn-engine -> longhorn-replica via TLS only or mTLS?
- longhorn-replica <-> longhorn-replica via TLS only or mTLS? 

Probably need to check longhorn-manager PR later, but introducing TLS in 1.3.0 should be fine as the first phase.

shuo-wu: I am confused about this part as well. Why are these authentication/encryption-related files optional while the server name is required?

innobead: @joshimoo any update? After resolving all comments, we can merge this.

joshimoo: Let me try to clarify:
- caFile is optional since the default are the root ca's shipped by linux
   - Example case where this is important, if you have a valid certificate produced by a root CA for example one doesn't need to provide a CA file since it's already part of the root CA chain. for example [let's encrypt](https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/)
- certFile&keyfile are optional in the client case, since they are only required when client authentication is used (Mutual TLS)
   - Forwarding thinking I could envision a use case where a client wants to connect to regular TLS server without MTLS, example use case are health/metrics endpoint, etc
- peerName is optional too, but requires a little bit more SSL knowledge to understand what it's used for, I will leave some links for context / reading below
   - Since we connect against an IP address instead of a DNS name, no host name would be available so we provide a peerName to check that the server/client returned certificate is for a specific set of hosts.
   - [Practical explanation](https://github.com/golang/go/issues/7342#issuecomment-66091467) ,[RFC TLS SNI](https://datatracker.ietf.org/doc/html/rfc6066#section-3)

I generally tried to keep this simple and generic, if you feel like hard coding more of this or removing the flexibility would be preferred let me know.

pkg/util/grpcutil.go
joshimoo: You are correct to create a cert/key pair one needs to provide both files:
- providing both files -> success
- providing only 1 of the files will lead to an error return --> error
- Providing 0 of the files is a valid use case though, so that's why it's || instead of and. --> success

pkg/util/grpcutil.go
innobead: Thanks, it's good now, especially for the clarification of the optional caFile.

pkg/util/grpcutil.go
innobead: ref: https://github.com/longhorn/longhorn-instance-manager/pull/109/files#r864656548

PR #113, opened 2022-05-05
Fallback to non-tls connection if tls files not found

innobead:
Signed-off-by: David Ko <dko@suse.com>

https://github.com/longhorn/longhorn/issues/3922


c3y1huang:
LGTM

c3y1huang (approved):

PR #115, opened 2022-05-11
fix cli im client non tls fallback

joshimoo:
- fixed the cli client non tls fallback
- added cli tcp:// to the default endpoint (listen/url)
- fallback to tcp in case endpoint doesn't specify proto (client/server)
- added unit tests for parseEndpoint function

longhorn/longhorn#3966

### DIFF
--- a/app/cmd/process.go
+++ b/app/cmd/process.go
@@ -202,11 +202,15 @@ func getProcessManagerClient(c *cli.Context) (*client.ProcessManagerClient, erro
 	tlsDir := c.GlobalString("tls-dir")
 
 	if tlsDir != "" {
-		return client.NewProcessManagerClientWithTLS(url,
+		imClient, err := client.NewProcessManagerClientWithTLS(url,
 			filepath.Join(tlsDir, "ca.crt"),
 			filepath.Join(tlsDir, "tls.crt"),
 			filepath.Join(tlsDir, "tls.key"),
 			"longhorn-backend.longhorn-system")
+		if err == nil {
+			return imClient, err
+		}
+		logrus.WithError(err).Info("Falling back to non tls client")
 	}
 
 	return client.NewProcessManagerClient(url, nil)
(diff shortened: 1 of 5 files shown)


app/cmd/process.go
```diff
@@ -202,11 +202,15 @@ func getProcessManagerClient(c *cli.Context) (*client.ProcessManagerClient, erro
...
 
 	if tlsDir != "" {
-		return client.NewProcessManagerClientWithTLS(url,
+		imClient, err := client.NewProcessManagerClientWithTLS(url,
 			filepath.Join(tlsDir, "ca.crt"),
 			filepath.Join(tlsDir, "tls.crt"),
 			filepath.Join(tlsDir, "tls.key"),
 			"longhorn-backend.longhorn-system")
+		if err == nil {
+			return imClient, err
```
shuo-wu, line 211: NIT:
```
return imClient, nil
```

app/cmd/process.go
```diff
@@ -202,11 +202,15 @@ func getProcessManagerClient(c *cli.Context) (*client.ProcessManagerClient, erro
...
+		}
+		logrus.WithError(err).Info("Falling back to non tls client")
 	}
 
 	return client.NewProcessManagerClient(url, nil)
```
shuo-wu, line 216: NIT: Can we directly modify this function and `NewProcessManagerClientWithTLS` so that the caller does not need to handle the fallback by itself?

innobead (approved):
LGTM

app/cmd/process.go
innobead: Feel this is on purpose to make `NewProcessManagerClientWithTLS` and `NewProcessManagerClient` do their responsibilities precisely, but whether fallback or not leaves to the caller instead. 

This is more about responsibility segregation.

innobead:
@joshimoo @shuo-wu have some comments, after resolving, then we can merge this.

app/cmd/process.go
shuo-wu: Make sense. Thank you for your explanation.

innobead:
It seems all nit comments, so let's see if @joshimoo improves it later after rc1.

PR #116, opened 2022-05-16
grpc-proxy: fix backup

c3y1huang:
Coretest didn't pass. We need this block.

https://github.com/longhorn/longhorn-instance-manager/pull/105#discussion_r873563004


innobead (approved):
LGTM, same as the current engine backupstatus implementation, but this makes me curious why we need twice `FetchBackupStatus` calls? 

We can tackle the reason later, because it's the existing implementation, probably some reason. Also, this also means we need to make the shared part between cli and proxy for each volume function as discussed before, so we will not miss any part.

PR #117, opened 2022-05-19
Replace Ping with get gRPC connections state and fix NPE

c3y1huang:
https://github.com/longhorn/longhorn/issues/3967

https://github.com/longhorn/longhorn-manager/pull/1338


innobead (approved):
LGTM, but I also feel we need an interface for ProxyClient, so operating through interfaces is always as a good habit :D, but this can be later.

PR #120, opened 2022-06-01
grpc-proxy: use ctx w/timeout for gRPC proxy client

c3y1huang:
https://github.com/longhorn/longhorn/issues/4060

### DIFF
--- a/pkg/client/proxy.go
+++ b/pkg/client/proxy.go
@@ -15,6 +15,7 @@ import (
 	"github.com/longhorn/longhorn-instance-manager/pkg/meta"
 	"github.com/longhorn/longhorn-instance-manager/pkg/util"
 
+	eclient "github.com/longhorn/longhorn-engine/pkg/controller/client"
 	emeta "github.com/longhorn/longhorn-engine/pkg/meta"
 )
 
@@ -95,8 +96,13 @@ func NewProxyClient(ctx context.Context, ctxCancel context.CancelFunc, address s
 }
 
 const (
-	GRPCServiceTimeout = 3 * time.Minute
+	GRPCServiceTimeout = eclient.GRPCServiceTimeout * 2
 )
+
+func getContextWithGRPCTimeout(parent context.Context) context.Context {
+	ctx, _ := context.WithTimeout(parent, GRPCServiceTimeout)
+	return ctx
+}
 
 func (c *ProxyClient) getProxyErrorPrefix(destination string) string {
 	return fmt.Sprintf("proxyServer=%v destination=%v:", c.ServiceURL, destination)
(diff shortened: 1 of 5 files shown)


pkg/client/proxy.go
```diff
@@ -98,6 +98,11 @@ const (
 	GRPCServiceTimeout = 3 * time.Minute
```
PhanLe1010, line 98: Should we have the timeout longer than [the one used in the proxy server](https://github.com/longhorn/longhorn-engine/blob/a63f0e0514872221c05ce6e2153a9164fa426f24/pkg/controller/client/controller_client.go#L40) to avoid the backend operation timeout earlier than this. ref to [link](https://github.com/longhorn/longhorn-manager/pull/1348#discussion_r884454508)

c3y1huang: Thank you. Updated.

pkg/client/proxy.go
```diff
@@ -95,9 +96,14 @@ func NewProxyClient(ctx context.Context, ctxCancel context.CancelFunc, address s
 }
 
 const (
-	GRPCServiceTimeout = 3 * time.Minute
+	GRPCServiceTimeout = eclient.GRPCServiceTimeout * 2
 )
 
+func (c *ProxyClient) getContextWithGRPCTimeout() context.Context {
```
innobead, line 102: Can we change to a function instead of a method of ProxyClient? and change the name as below.

`createContextForProxyClientGRPCTimeout`

innobead (changes requested):

pkg/client/proxy.go
c3y1huang: Updated.

innobead (approved):

pkg/client/proxy_replica.go
```diff
@@ -33,7 +33,7 @@ func (c *ProxyClient) ReplicaAdd(serviceAddress, replicaAddress string, restore
 		ReplicaAddress: replicaAddress,
 		Restore:        restore,
 	}
-	_, err = c.service.ReplicaAdd(c.ctx, req)
+	_, err = c.service.ReplicaAdd(getContextWithGRPCTimeout(c.ctx), req)
```
shuo-wu, line 36: One important question: Is this timeout large enough? IIRC, this may take up to 24 hours.

https://github.com/longhorn/longhorn-instance-manager/blob/a304db671f4fd4ab456efab0e9fba1897a6dded5/pkg/proxy/replica.go#L27-L35

pkg/client/proxy_snapshot.go
```diff
@@ -121,7 +121,7 @@ func (c *ProxyClient) SnapshotClone(serviceAddress, name, fromController string)
 		SnapshotName:              name,
 		ExportBackingImageIfExist: false,
 	}
-	_, err = c.service.SnapshotClone(c.ctx, req)
+	_, err = c.service.SnapshotClone(getContextWithGRPCTimeout(c.ctx), req)
```
shuo-wu, line 124: Similar here.

pkg/client/proxy_replica.go
c3y1huang: https://github.com/longhorn/longhorn/issues/4096

PR #121, opened 2022-06-01
Fix incorrect error returned of BackupRestore()

derekbit:
[Longhorn 4049](https://github.com/longhorn/longhorn/issues/4049)

### DIFF
--- a/pkg/client/proxy_backup.go
+++ b/pkg/client/proxy_backup.go
@@ -94,6 +94,10 @@ func (c *ProxyClient) BackupRestore(serviceAddress, url, target, volumeName stri
 	}
 
 	defer func() {
+		if _, ok := err.(TaskError); ok {
+			return
+		}
+
 		err = errors.Wrapf(err, "%v failed to restore backup %v to volume %v", c.getProxyErrorPrefix(serviceAddress), url, volumeName)
 	}()
 


pkg/client/proxy_backup.go
```diff
@@ -160,7 +154,6 @@ func (c *ProxyClient) BackupRestoreStatus(serviceAddress string) (status map[str
 	}
 	return status, nil
 }
-
```
c3y1huang, line 163: ?

c3y1huang (changes requested):

pkg/client/proxy_backup.go
derekbit: Updated. Thank you.

c3y1huang (approved):
lgtm

pkg/client/proxy_backup.go
```diff
@@ -108,18 +104,16 @@ func (c *ProxyClient) BackupRestore(serviceAddress, url, target, volumeName stri
 	}
 	recv, err := c.service.BackupRestore(c.ctx, req)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "%v failed to restore backup %v to volume %v", c.getProxyErrorPrefix(serviceAddress), url, volumeName)
```
innobead, line 107: Should we just keep the original defer function to make sure we have a consistent error format? Also found other methods in this file having the same issues?

```golang
	defer func() {
	    if err != nil {
		    err = errors.Wrapf(err, "%v failed to restore backup %v to volume %v", c.getProxyErrorPrefix(serviceAddress), url, volumeName)
		}
	}()
```

innobead (changes requested):

c3y1huang (commented):
nit: don't need to check if err is nil in defer.
https://github.com/longhorn/longhorn-instance-manager/blob/master/vendor/github.com/pkg/errors/errors.go#L202

pkg/client/proxy_backup.go
derekbit: Yes. Addressed!

innobead:
> nit: don't need to check if err is nil in defer. It's checked in the `errors.Wrapf` [https://github.com/longhorn/longhorn-instance-manager/blob/master/vendor/github.com/pkg/errors/errors.go#L202](https://github.com/longhorn/longhorn-instance-manager/blob/master/vendor/github.com/pkg/errors/errors.go?rgh-link-date=2022-06-02T05%3A45%3A40Z#L202)

Didn't notice this implicit. Thanks, @c3y1huang !

pkg/client/proxy_backup.go
```diff
@@ -94,7 +94,12 @@ func (c *ProxyClient) BackupRestore(serviceAddress, url, target, volumeName stri
 	}
 
 	defer func() {
-		err = errors.Wrapf(err, "%v failed to restore backup %v to volume %v", c.getProxyErrorPrefix(serviceAddress), url, volumeName)
+		if err != nil {
+			_, ok := err.(TaskError)
```
innobead, line 98: Finally know why you are doing this. 

Because the caller of this method will check the concert type of the returned error, you need to return the concrete one instead of another wrapped error.

innobead (changes requested):

derekbit:
> @derekbit Could you help with updating manifests of longhorn-manager and longhorn repo? Thank you.
> 
> https://github.com/longhorn/longhorn-instance-manager/releases/tag/v1_20220602

No prob. Addressed!

PR #124, opened 2022-06-09
grpc-proxy: use long timeout for proxy clients

c3y1huang:
https://github.com/longhorn/longhorn/issues/4096

### DIFF
--- a/pkg/client/proxy.go
+++ b/pkg/client/proxy.go
@@ -15,8 +15,8 @@ import (
 	"github.com/longhorn/longhorn-instance-manager/pkg/meta"
 	"github.com/longhorn/longhorn-instance-manager/pkg/util"
 
-	eclient "github.com/longhorn/longhorn-engine/pkg/controller/client"
 	emeta "github.com/longhorn/longhorn-engine/pkg/meta"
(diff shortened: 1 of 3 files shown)


pkg/client/proxy.go
```diff
@@ -96,14 +96,20 @@ func NewProxyClient(ctx context.Context, ctxCancel context.CancelFunc, address s
 }
 
 const (
-	GRPCServiceTimeout = eclient.GRPCServiceTimeout * 2
+	GRPCServiceTimeout     = eclient.GRPCServiceCommonTimeout * 2
+	GRPCServiceLongTimeout = eclient.GRPCServiceLongTimeout * 2
```
shuo-wu, line 100: I don't think we need to doubt the long timeout. Instead, this can be `GRPCServiceLongTimeout = eclient.GRPCServiceLongTimeout + GRPCServiceTimeout`

shuo-wu (changes requested):
As I mentioned in the ticket, there is no need to set a super long timeout for some APIs.

innobead (approved):

shuo-wu (approved):

PR #126, opened 2022-06-24
Add tcmalloc for amd64 platform

derekbit:
Depends on https://github.com/rancher/tgt/pull/8 and https://github.com/rancher/liblonghorn/pull/16

[Longhorn 4136](https://github.com/longhorn/longhorn/issues/4136)

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -12,6 +12,10 @@ RUN zypper -n install cmake kmod curl nfs-client nfs4-acl-tools fuse git gcc \
     iputils telnet iperf qemu-tools wget iproute2 xsltproc docbook-xsl-stylesheets \
     e2fsprogs e2fsprogs-devel xfsprogs xfsprogs-devel
 
+RUN if [ ${ARCH} == "amd64" ]; then \
+        zypper -n install autoconf libtool libunwind-devel gcc-c++; \
+    fi
+
 # Install grpc_health_probe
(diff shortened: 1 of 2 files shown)


package/Dockerfile
```diff
@@ -20,23 +24,46 @@ COPY bin/longhorn-instance-manager /usr/local/bin/
...
+RUN if [ ${ARCH} == "amd64" ]; then \
+        cd /usr/src && \
+        git clone https://github.com/gperftools/gperftools.git&& \
+        cd gperftools && \
+        git checkout db6d94d9893fdd8c74028b5113cd9185be1e3111 && \
+        ./autogen.sh && \
+        ./configure && \
+        make && \
+        make install; \
+    fi
```
shuo-wu, line 40: Is `rm -rf /usr/src/gperftools` missed here?
Similar below

shuo-wu (commented):
Do we have a ticket to track the arm64 part?

package/Dockerfile
derekbit: Yeah, we can delete it.

derekbit:
> Do we have a ticket to track the arm64 part?

No, need to investigate the arm64 part later.
I will create a ticket tracking arm64 part.

joshimoo (approved):
looks okay to me. 

Seems like a hack though relying on https://github.com/gperftools/gperftools.git installation just to get tcmalloc instead of building directly via https://github.com/google/tcmalloc but setting up bazel etc could be tedious.

Let's see if gperftools also has something for ARM64 otherwise we should go with relying/building tcmalloc directly.

package/Dockerfile
derekbit: Addressed. Thank you.

innobead (approved):

PR #129, opened 2022-08-01
fix: none-json error from backup store

c3y1huang:
Signed-off-by: Chin-Ya Huang <chin-ya.huang@suse.com>

### DIFF
--- a/pkg/client/proxy_backup.go
+++ b/pkg/client/proxy_backup.go
@@ -115,7 +115,7 @@ func (c *ProxyClient) BackupRestore(serviceAddress, url, target, volumeName stri
 	if recv.TaskError != nil {
 		var taskErr TaskError
 		if jsonErr := json.Unmarshal(recv.TaskError, &taskErr); jsonErr != nil {
-			err = errors.Wrap(jsonErr, "Cannot unmarshal the restore error, maybe it's not caused by the replica restore failure")
+			err = errors.Wrapf(jsonErr, "cannot unmarshal the restore error, maybe it's not caused by the replica restore failure: %s", recv.TaskError)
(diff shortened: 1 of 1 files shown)


derekbit (approved):

pkg/client/proxy_backup.go
```diff
@@ -115,7 +116,7 @@ func (c *ProxyClient) BackupRestore(serviceAddress, url, target, volumeName stri
 	if recv.TaskError != nil {
 		var taskErr TaskError
 		if jsonErr := json.Unmarshal(recv.TaskError, &taskErr); jsonErr != nil {
-			err = errors.Wrap(jsonErr, "Cannot unmarshal the restore error, maybe it's not caused by the replica restore failure")
+			err = errors.Wrap(fmt.Errorf(string(recv.TaskError)), "Cannot unmarshal the restore error, maybe it's not caused by the replica restore failure")
```
shuo-wu, line 119: `TaskError` already implements the interface `Error()`. I think you can directly use it `errors.Wrap(recv.TaskError, ......)`.
BTW, should we output `jsonErr` either?

shuo-wu (changes requested):

pkg/client/proxy_backup.go
innobead: We should, even though the decode error is just a cascading error, but it will help us know the complete error context.

c3y1huang: Updated.

innobead (approved):

innobead:
@c3y1huang Let's backport this to 1.3.x. Thanks.

PR #134, opened 2022-09-06
Tagging netgo to be used while compiling with cgo 

olljanat:
~Implements suggestion in https://github.com/longhorn/longhorn/issues/4415#issuecomment-1214885963 and disable CGO here too like it was done in longhorn-manager by https://github.com/longhorn/longhorn-manager/pull/61~

Looks that instance-manager needs CGO so switched to alternative solution proposed in https://github.com/longhorn/longhorn/issues/1768#issuecomment-690778513


scripts/build
```diff
@@ -6,7 +6,7 @@ source $(dirname $0)/version
 cd $(dirname $0)/..
 
 mkdir -p bin
-go build -ldflags \
+go build --tags netgo -ldflags \
```
innobead, line 9:
```suggestion
go build -tags netgo -ldflags \
```

innobead (changes requested):

scripts/build
olljanat: Good catch. Fixed.

olljanat:
@innobead typo fixed, please check again.

olljanat:
PR is still valid

innobead:
Because the instance manager has the proxy capability and it is actually the core of the data plane, we need this as well instead of longhorn-engine only. (probably we will reach out to the service in DNS from the instance manager)

innobead (approved):
Merge after QA verification. 

cc @longhorn/qa @roger-ryao

roger-ryao:
> Merge after QA verification.
> 
> cc @longhorn/qa @roger-ryao

Hi @innobead 
- [core test report](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2448/)
- [regression test report](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2447/)
- [The failed cases re-run report](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2457/)

The test PASS, we could merge longhorn-instance-manager PR
cc @longhorn/qa

PR #137, opened 2022-09-28
build: update bci-base image to 15.4

yangchiu:
build: update bci-base image to 15.4

For https://github.com/longhorn/longhorn/issues/4617

Signed-off-by: Yang Chiu <yang.chiu@suse.com>

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,4 +1,4 @@
-FROM registry.suse.com/bci/bci-base:15.3
+FROM registry.suse.com/bci/bci-base:15.4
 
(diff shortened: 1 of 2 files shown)


yangchiu:
arm64 and s390x build failed because they're not found in https://download.opensuse.org/repositories/network:/utilities/15.4/

yangchiu:
> arm64 and s390x build failed because they're not found in https://download.opensuse.org/repositories/network:/utilities/15.4/

https://download.opensuse.org/repositories/network:/utilities/SLE_15_SP4/ already updated.

PR #139, opened 2022-10-04
Adjust the iSCSI initiator/target timeout and the engine-to-replica timeout

derekbit:
https://github.com/longhorn/longhorn/issues/4491


derekbit:
Will update the dependencies if https://github.com/longhorn/longhorn-engine/pull/762 is merged.

derekbit:
The PR is not really necessary.

innobead:
@derekbit Why not required? 

Is it because the chance is related to starting engine process call with general arguments provided by longhorn manager, so nothings required to change here but longhorn manager instead.

derekbit:
The engine process create/upgrade still rely on the longhorn-engine binary's command line, so the update of the instance-manager is necessary for this feature.

innobead:
Exactly, clear.

derekbit:
BTW, alway updating the instance-manager to the latest might be a good way? Let me reopen it. WDYT?

innobead:
Sounds good to me.

derekbit:
Will update the dependencies if https://github.com/longhorn/longhorn-engine/pull/762 is merged

innobead (approved):

PR #140, opened 2022-10-11
feat: Snapshot checksum calculation and verification

derekbit:
https://github.com/longhorn/longhorn/issues/4210
https://github.com/longhorn/longhorn/issues/3198


pkg/client/proxy_snapshot.go
```diff
@@ -5,12 +5,10 @@ import (
 
 	etypes "github.com/longhorn/longhorn-engine/pkg/types"
 	eutil "github.com/longhorn/longhorn-engine/pkg/util"
+	ediskutil "github.com/longhorn/longhorn-engine/pkg/util/disk"
```
shuo-wu, line 8: Move const `VolumeHeadName` to another file like `"github.com/longhorn/longhorn-engine/pkg/types/types"` then we can use alias `enginetypes`

pkg/imrpc/proxy.proto
```diff
@@ -214,6 +216,13 @@ message EngineReplicaRemoveRequest {
 	string replica_address = 2;
 }
 
+message EngineReplicaUpdateRequest {
```
shuo-wu, line 219: 1. This is more like `ReplicaModeUpdate`
2. The field `mode` is a controller in-memory field. Should replicas understand it?

pkg/client/proxy_snapshot.go
derekbit: Sound good. Thank you.

pkg/imrpc/proxy.proto
derekbit: Updated. Thank you.

derekbit:
Please help review. Thank you.
cc @shuo-wu @c3y1huang @innobead

shuo-wu (dismissed):
LGTM if there is no doubt about the engine part

derekbit:
Will update dependencies if longhorn-engine PR is merged.

derekbit:
This is ready for review.
cc @innobead

innobead (approved):

PR #141, opened 2022-11-02
Filesystem Trim Support

shuo-wu:
longhorn/longhorn#836


innobead (approved):
LGTM, merge this until the engine dependency update

generate_grpc.sh
```diff
@@ -14,7 +14,7 @@ if [ ! -e ${DIR} ]; then
     rm -rf ${DIR}
     mkdir -p $DIR
     wget https://raw.githubusercontent.com/longhorn/longhorn-engine/master/proto/ptypes/common.proto -P $DIR
-    wget https://raw.githubusercontent.com/longhorn/longhorn-engine/master/proto/ptypes/controller.proto -P $DIR
+    wget https://raw.githubusercontent.com/shuo-wu/longhorn-engine/trim-dev/proto/ptypes/controller.proto -P $DIR
```
innobead, line 17: This needs to change.

innobead (changes requested):

generate_grpc.sh
```diff
@@ -14,7 +14,7 @@ if [ ! -e ${DIR} ]; then
     rm -rf ${DIR}
     mkdir -p $DIR
     wget https://raw.githubusercontent.com/longhorn/longhorn-engine/master/proto/ptypes/common.proto -P $DIR
-    wget https://raw.githubusercontent.com/longhorn/longhorn-engine/master/proto/ptypes/controller.proto -P $DIR
+    wget https://raw.githubusercontent.com/longhorn/longhorn-engine/trim-dev/proto/ptypes/controller.proto -P $DIR
```
innobead, line 17: This is not found `https://raw.githubusercontent.com/longhorn/longhorn-engine/trim-dev/proto/ptypes/controller.proto`, do we need to change this? I think it should be kept at master

innobead (changes requested):

generate_grpc.sh
shuo-wu: My fault... Should be `master`.

innobead (approved):

PR #143, opened 2022-11-28
Dockerfile: update the path to devel tools repo

derekbit:
[Longhorn 4946](https://github.com/longhorn/longhorn/issues/4946)


c3y1huang (approved):

c3y1huang:
Note: we also need to fix the other repos. 
Ref: https://github.com/longhorn/longhorn/issues/4617

innobead (approved):

innobead:
> Note: we also need to fix the other repos. Ref: [longhorn/longhorn#4617](https://github.com/longhorn/longhorn/issues/4617)
> 
> [longhorn/longhorn#4949](https://github.com/longhorn/longhorn/issues/4949)

@yangchiu Could you help with this? Thanks

PR #144, opened 2022-11-28
Rebuilding speedup: update proto and proxy apis

derekbit:
[Longhorn 4783](https://github.com/longhorn/longhorn/issues/4783)


derekbit:
Dependeny is updated.

innobead (approved):

PR #145, opened 2022-12-09
Change some log messages to trace level

derekbit:
[Longhorn 4427
](https://github.com/longhorn/longhorn/issues/4427)


innobead (approved):
LGTM.

We don't use log level correctly, at least from log level definition.

Most logs should be info for intended info but some tracing purpose (verbose log) should debug or even trace.  However, we did everything in debug. 

This fix is a good start for log level management!

cc @longhorn/dev

derekbit:
https://github.com/Mergifyio backport v1.3.x

PR #158, opened 2022-12-15
Support metrics for IOPS and bandwidth real time

derekbit:
https://github.com/longhorn/longhorn/issues/2406

Depends on https://github.com/longhorn/longhorn-engine/pull/794


pkg/client/proxy_metrics.go
```diff
@@ -0,0 +1,37 @@
...
+
+	req := &rpc.ProxyEngineRequest{
+		Address: serviceAddress,
+	}
+	resp, err := c.service.MetricsGet(getContextWithGRPCTimeout(c.ctx), req)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Metrics{
```
innobead, line 29: It seems like we have similar types defined in different components like manager, instance-manager, engine or even replica. It's better to consolidate them in one place via protobuf, then reused together. A data model repo.  

Let's create a ticket for this refactoring.

innobead (dismissed):
LGTM.

innobead (approved):

innobead (approved):

PR #171, opened 2023-01-03
Improve backup performance with faster compression algorithm and mutilple threads

derekbit:
https://github.com/longhorn/longhorn/issues/5189


innobead (changes requested):
Please update the dep of backupstore again, because there is a update there.

derekbit:
Dependencies are updated.

innobead (approved):
LGTM

shuo-wu:
Should we update the API version for engine or instance manager?

derekbit:
> Should we update the API version for engine or instance manager?

Yeah, we should update the API versions.
Will create a ticket for it.

PR #175, opened 2023-01-16
[BACKPORT][1.4.x] vendor: update backupstore

mantissahz:
update to v0.0.0-20230125201126-0c2bd550ebc3

longhorn/longhorn#5266


innobead (dismissed):

mantissahz:
Need to update new fixed.

innobead:
@mantissahz having separate commits should be meaningful. Here these two commits have the same purpose, so you should force commit as one message. (even though we can do the squash when merging the PR, but this is a good habit to get used to)

image

innobead (approved):

PR #178, opened 2023-02-08
build: remove devel:tools:scm repo

yangchiu:
build: remove devel:tools:scm repo

For https://github.com/longhorn/longhorn/issues/5365

Signed-off-by: Yang Chiu <yang.chiu@suse.com>


innobead (approved):
LGTM

PR #186, opened 2023-02-16
Clean up unused mount points

derekbit:
https://github.com/longhorn/longhorn/issues/5391

deps on
https://github.com/longhorn/backupstore/pull/102
https://github.com/longhorn/longhorn-engine/pull/854


derekbit:
Handled in https://github.com/longhorn/longhorn-instance-manager/pull/188

PR #198, opened 2023-03-14
vendor: update backupstore

mantissahz:
Update `backupstore` from `v0.0.0-20220913112826-5f5c95274f2a` to `v0.0.0-20230125201126-0c2bd550ebc3` for proxy client setting.

longhorn/longhorn#5265

Backport accidentally https://github.com/longhorn/longhorn/issues/5265#issuecomment-1467794567


innobead:
@derekbit would this cause some compatibility issues? IIRC, some interface changed in backupstore so we have 1.4.x, but didn't have 1.3.

innobead:
@mantissahz where the commit is from? from the master branch?

derekbit:
@mantissahz @innobead 
I think we should create a branch v1.3.x in backupstore. WDYT?

innobead:
Yeah, but the reason needs to clarify first.

We probably need to explain why we need to have branches somehow for backupstore and spare tool.

As discussed, technically we should make common libs independent and decoupled if possible.(need to consider from 1.5)

mantissahz:
> where the commit is from? from the master branch?

Yes, it's from master branch.

> would this cause some compatibility issues?

I have concerns about the compatibility as well

> we should create a branch v1.3.x in backupstore

Because we have v1.4.x as well

derekbit:
The backupstore introduces new features and fields for backup compression enhancement which is not in v1.3.x and v1.4.x. 
That's why we need branches for different versions for now.

innobead:
@mantissahz 
If having concerns, it's better to raise first or make the PR draf; otherwise it would be possibly merged accidentally.

Let's have another branch for 1.3, I will update readme of backupstore and spare tool to avoid some confusion.

innobead:
@mantissahz Please use the code from https://github.com/longhorn/backupstore/tree/v1.3.x instead. Also, help update this dependency to longhorn-manager, longhorn-instance-manager, and longhorn-engine.

cc @derekbit 

P.S. https://github.com/longhorn/backupstore/tree/v1.3.x has system-backup/restore implementation which is for 1.4, but it's implementation is compatible, so that is fine to include in the 1.3.x branch.

PR #208, opened 2023-04-06
Fix bug: volume is stuck in live engine upgrading

PhanLe1010:
The only usecase of ProcessReplace is for engine image live upgrade. If the new engine process has the same binary as the existing engine process, there is no reason to do the ProcessReplace

longhorn/longhorn#5684


innobead (approved):

innobead:
@Mergifyio backport v1.4.x v1.3.x

PR #213, opened 2023-04-14
vendor: update longhorn-engine

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/4305

`sparse-tool` is updated so the `longhorn-engine` is updated.
This PR updates the longhorn-engine in the longhorn-instance-manager


innobead (approved):

ChanYiLin:
Real backport PR: https://github.com/longhorn/longhorn-instance-manager/pull/215

PR #216, opened 2023-04-19
spdk: build `spdk` and update engine-manager

derekbit:


derekbit:
cc @shuo-wu

derekbit:
Will check the failure on `arm64` platform later.

derekbit:
Handled in https://github.com/longhorn/longhorn-instance-manager/pull/218/commits.

PR #218, opened 2023-04-30
Make longhorn-instance-manager support SPDK features

derekbit:


pkg/disk/disk.go
```diff
@@ -0,0 +1,354 @@
...
+
+	spdkClient, err := s.newSPDKClient()
+	if err != nil {
+		return nil, grpcstatus.Error(grpccodes.Internal, errors.Wrap(err, "failed to get SPDK client").Error())
+	}
+	defer spdkClient.Close()
+
+	diskPath := getDiskPath(req.DiskPath)
+
+	bdevName, err := spdkClient.BdevAioCreate(diskPath, req.DiskName, blockSize)
```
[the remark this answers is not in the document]
shuo-wu: And the following part is duplicated with `DiskGet()`

pkg/disk/error.go
```diff
@@ -0,0 +1,41 @@
...
+	"syscall"
+)
+
+type ErrorMessage struct {
+	Code    int    `json:"code"`
+	Message string `json:"message"`
+}
+
+func parseErrorMessage(errStr string) (*ErrorMessage, error) {
+	r := regexp.MustCompile(`"code": (-?\d+),\n\t"message": "([^"]+)"`)
```
shuo-wu, line 16: NIT: Make the regex regex a const

pkg/disk/error.go
```diff
@@ -0,0 +1,41 @@
...
+
+	em := &ErrorMessage{
+		Code:    code,
+		Message: match[2],
+	}
+
+	return em, nil
+}
+
+func isFileExists(message string) bool {
```
shuo-wu, line 35: Quesion: Should we make some spdk APIs idempodent? For example, DiskCreate, DiskDelete, Expose...

pkg/disk/disk.go
```diff
@@ -0,0 +1,354 @@
...
+		Uuid:        lvstoreInfo.UUID,
+		Path:        req.DiskPath,
+		Type:        DiskTypeBlock,
+		TotalSize:   int64(lvstoreInfo.TotalDataClusters * lvstoreInfo.ClusterSize),
+		FreeSize:    int64(lvstoreInfo.FreeClusters * lvstoreInfo.ClusterSize),
+		TotalBlocks: int64(lvstoreInfo.TotalDataClusters * lvstoreInfo.ClusterSize / lvstoreInfo.BlockSize),
+		FreeBlocks:  int64(lvstoreInfo.FreeClusters * lvstoreInfo.ClusterSize / lvstoreInfo.BlockSize),
+		BlockSize:   int64(lvstoreInfo.BlockSize),
+		ClusterSize: int64(lvstoreInfo.ClusterSize),
+		Readonly:    false,
```
shuo-wu, line 214: Is this necessary (at least for now)?

pkg/imrpc/instance.proto
```diff
@@ -0,0 +1,94 @@
...
+
+message ProcessSpecific {
+	string binary = 1;
+	repeated string args = 2;
+}
+
+message SpdkSpecific {
+	string address = 1;
+    map<string, string> replica_address_map = 2;
+    string frontend = 3;
```
shuo-wu, line 28: NIT: Wrong indent

pkg/imrpc/instance.proto
```diff
@@ -0,0 +1,94 @@
...
+message InstanceStatus {
+	string state = 1;
+	string error_msg = 2;
+	int32 port_start = 3;
+	int32 port_end = 4;
+}
+
+message InstanceCreateRequest {
+	InstanceSpec spec = 1;
+	bool expose_required = 2;
```
shuo-wu, line 55: The caller of this API should not care about this. The Instance or SPDK engine gRPC server can determine that by itself.

pkg/imrpc/instance.proto
```diff
@@ -0,0 +1,94 @@
...
+}
+
+message InstanceDeleteRequest {
+	string name = 1;
+	string type = 2;
+	string backend_store_driver = 3;
+
+	string disk_uuid = 4;
```
shuo-wu, line 63: Does the caller of this API need to provide this? The instance name should be unique and the gRPC server can figure out the disk by itself.

pkg/imrpc/instance.proto
```diff
@@ -0,0 +1,94 @@
...
+	bool cleanup_required = 5;
+}
+
+message InstanceGetRequest {
+	string name = 1;
+	string type = 2;
+	string backend_store_driver = 3;
+
+	string disk_uuid = 4;
```
shuo-wu, line 72: the same here

pkg/instance/instance.go
```diff
@@ -0,0 +1,588 @@
...
+	grpcstatus "google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/emptypb"
+
+	"github.com/longhorn/longhorn-instance-manager/pkg/client"
+	rpc "github.com/longhorn/longhorn-instance-manager/pkg/imrpc"
+	"github.com/longhorn/longhorn-instance-manager/pkg/meta"
+	"github.com/longhorn/longhorn-instance-manager/pkg/types"
+)
+
+type Server struct {
```
shuo-wu, line 22: Cache `ProcessManagerClient` here.
BTW, does `ProcessManagerClient` need to care about TLS now?

pkg/imrpc/instance.proto
shuo-wu: I am thinking if there is a better way to input the necessary args for engines or replicas.

pkg/spdk/spdk.go
```diff
@@ -0,0 +1,724 @@
+package spdk
```
shuo-wu, line 1: I will skip the review of this commit since we will use the implementation in `longhorn-spdk-engine` in the end.

shuo-wu:
Some of the comments are related to the SPDK engine service APIs. You can update them later.

package/Dockerfile
```diff
@@ -39,6 +41,18 @@ RUN cd /usr/src && \
...
+ENV SPDK_DIR /usr/src/spdk
+RUN git clone https://github.com/derekbit/spdk.git ${SPDK_DIR} --recursive && \
+    cd ${SPDK_DIR} && \
+    git checkout 83584882eecf1837a4e0cc632d730e1ff6c8d888 && \
+    ./scripts/pkgdep.sh && \
+    git submodule update --init && \
+    ./configure --disable-tests --disable-unit-tests --disable-examples && \
+    make -j$(nproc) && \
+    make install && \
+    rm -rf /usr/src/spdk
```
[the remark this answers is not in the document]
derekbit: `spdk_tgt` is installed to `/usr/local/bin`.
Do you prefer to keep the source code?

pkg/disk/disk.go
```diff
@@ -0,0 +1,354 @@
...
+	}, nil
+}
+
+func (s *Server) bdevLvolGetLvstore(spdkClient *spdkclient.Client, lvsName, uuid string) (*spdktypes.LvstoreInfo, error) {
```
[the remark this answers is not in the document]
derekbit: I agree moving this helper function into `go-spdk-helper`. 
I will create a PR for this part.

pkg/disk/disk.go
```diff
@@ -0,0 +1,354 @@
...
+
+	// A disk does not have a fsid, so we use the device number as the disk ID
+	diskID, err := getDeviceNumber(diskPath)
+	if err != nil {
+		log.WithError(err).Error("Failed to get disk ID")
+		return nil, grpcstatus.Error(grpccodes.Internal, errors.Wrapf(err, "failed to get disk ID").Error())
+	}
+
+	return &rpc.Disk{
+		Id:          diskID,
```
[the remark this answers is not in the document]
derekbit: Actually, I'm not comfortable with the major-minor number as an ID. I'm thinking if we can let ID is same as UUID.
`ID` is `fsid` for filesystem-type disk, so it is very unique. For a block-type disk, UUID might be better. WDYT?

pkg/disk/error.go
derekbit: Can you elaborate more on `idempodent` ?

pkg/imrpc/instance.proto
derekbit: Agree. Current layout is a bit messy.

pkg/imrpc/instance.proto
derekbit: I see. 
So, the exposure of replicas will happen while creating an engine?

pkg/imrpc/instance.proto
derekbit: The name of a lvols is `diskUUID/replica-name`.
If the diskUUID is not provided, will the SPDK service need to iterate the lvols?

pkg/instance/instance.go
derekbit: Do you mean we don't care about TLS for ProcessManagerClient because it is an internal service?

pkg/disk/disk.go
```diff
@@ -0,0 +1,354 @@
...
+	}
+
+	spdkClient, err := s.newSPDKClient()
+	if err != nil {
+		return nil, grpcstatus.Error(grpccodes.Internal, errors.Wrapf(err, "failed to get SPDK client").Error())
+	}
+	defer spdkClient.Close()
+
+	// Check if the disk exists
+	bdevInfos, err := spdkClient.BdevGetBdevs(req.DiskName, 3000)
```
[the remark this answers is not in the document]
derekbit: BTW, do you have any suggestion for the timeout?

pkg/imrpc/instance.proto
derekbit: Can we add `replica_specific` and `engine_specific` structs? WDYT?

pkg/instance/instance.go
shuo-wu: Yes.

pkg/imrpc/instance.proto
shuo-wu: Wait... Maybe retaining this would be better...

package/Dockerfile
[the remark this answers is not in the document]
shuo-wu: Then please use `${SPDK_DIR}` in this line.

pkg/disk/disk.go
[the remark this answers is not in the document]
shuo-wu: 1. We can store the disk name (which is provided by users and is the lvs-name for SPDK) in a new field `name`
2. As for this field `ID`, it can be the major-minor for SPDK/block disk and fsid for ds disk. I don't think we need to make this the same as `UUID` as we already have that field.

pkg/disk/disk.go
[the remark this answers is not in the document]
shuo-wu: Sure. You can add more API functions in `go-spdk-helper/pkg/spdk/client/basic.go`. Thank you!

pkg/disk/error.go
shuo-wu: Deleting a non-existing disk always returns `nil` rather than an error.
Stopping exposing a stopped replica Lvol always returns `nil`.

pkg/imrpc/instance.proto
shuo-wu: Actually, another way to handle this is similar to backing image creation/the process starting: Use a string map to store all necessary input fields. Then there won't be an empty value for fields like `frontend` in replica `InstanceCreate` requests. But the defect is, new developers do not know which fields are necessary by checking the API definition.
I am fine with either. WDYT?

pkg/disk/disk.go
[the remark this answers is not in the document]
derekbit: I see. Let's use major-minor. How about it?

package/Dockerfile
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

package/instance-manager
```diff
@@ -1,7 +1,60 @@
...
+
+function enable_tgtd() {
+    echo "Enabling tgtd"
+    tgtd -f 2>&1 | tee /var/log/tgtd.log &
+}
+
+function enable_spdk_tgt() {
+    local options=("$@")
+    echo "Enabling spdk_tgt with options: ${options[*]}"
+    spdk_tgt "${options[@]}" 2>&1 | tee /var/log/spdk_tgt.log &
```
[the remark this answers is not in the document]
derekbit: What do we need this?
BTW, do we need liveness probes for `spdk_tgt` and `tgtd`?

pkg/client/proxy_backup.go
```diff
@@ -8,11 +8,13 @@ import (
 	rpc "github.com/longhorn/longhorn-instance-manager/pkg/imrpc"
 )
 
-func (c *ProxyClient) SnapshotBackup(serviceAddress, backupName, snapshotName, backupTarget,
+func (c *ProxyClient) SnapshotBackup(engineName, serviceAddress, backendStoreDriver, backupName, snapshotName, backupTarget,
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

pkg/imrpc/proxy.proto
```diff
@@ -43,6 +43,8 @@ service ProxyEngineService {
 
 message ProxyEngineRequest{
 	string address = 1;
+	string engine_name = 2;
+	string backend_store_driver = 3;
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

pkg/proxy/replica.go
```diff
@@ -73,10 +92,67 @@ func (p *Proxy) ReplicaList(ctx context.Context, req *rpc.ProxyEngineRequest) (r
...
+	case rpc.ReplicaMode_RW:
+		return eptypes.ReplicaMode_RW
+	case rpc.ReplicaMode_ERR:
+		return eptypes.ReplicaMode_ERR
+	}
+	return eptypes.ReplicaMode_ERR
+}
+
+func (p *Proxy) replicaListFromSpdkService(ctx context.Context, req *rpc.ProxyEngineRequest) (resp *rpc.EngineReplicaListProxyResponse, err error) {
+	c, err := client.NewSPDKServiceClient("tcp://"+p.spdkServiceAddress, nil)
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

pkg/client/disk_types.go
```diff
@@ -0,0 +1,15 @@
+package client
+
+type DiskInfo struct {
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

pkg/disk/disk.go
```diff
@@ -0,0 +1,354 @@
...
+		shutdownCh:    shutdownCh,
+		HealthChecker: &GRPCHealthChecker{},
+	}
+
+	go s.startMonitoring()
+
+	return s, nil
+}
+
+func (s *Server) newSPDKClient() (*spdkclient.Client, error) {
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

pkg/disk/disk.go
```diff
@@ -0,0 +1,354 @@
...
+		return nil, grpcstatus.Error(grpccodes.InvalidArgument, "disk name and disk path are required")
+	}
+
+	blockSize := uint64(defaultBlockSize)
+	if req.BlockSize > 0 {
+		log.Infof("Using custom block size %v", req.BlockSize)
+		blockSize = uint64(req.BlockSize)
+	}
+
+	ok, err := isBlockDevice(req.DiskPath)
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

pkg/disk/disk.go
derekbit: Addressed. Thank you.

pkg/imrpc/instance.proto
derekbit: OK!

pkg/imrpc/instance.proto
derekbit: How about add the tow maps `replica_specific map[string]string` and `engine_specific map[string]string`?

pkg/disk/disk.go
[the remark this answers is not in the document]
shuo-wu: NO for now. I guess 1 min may be enough for now

pkg/api/types.go
```diff
@@ -196,3 +196,12 @@ type DiskInfo struct {
 	BlockSize   int64
 	ClusterSize int64
 }
+
+type ReplicaInstance struct {
```
shuo-wu, line 200: Why can't we directly use `Replica` in `longhorn-spdk-engine/pkg/api`?
I am thinking if there are too many replica/engine related struct definitions/structs...

pkg/imrpc/imcommon.proto
```diff
@@ -0,0 +1,30 @@
...
+
+    int64 instanceManagerInstanceServiceAPIVersion = 10;
+    int64 instanceManagerInstanceServiceAPIMinVersion =11;
+}
+
+message LogResponse {
+	string line = 2;
+}
+
+enum BackendStoreDriver {
```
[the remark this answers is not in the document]
derekbit: Could you elaborate more on it? TBH, I don't see the obvious benefit of the change...

package/instance-manager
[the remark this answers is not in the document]
derekbit: No, need to monitor them in the instance manager process. I don't want to monitor them for now. If we automatically restart the IM pod, the spdk_tgt will be missing. We can improve this part after finishing the log collection as we discussed before.

pkg/api/types.go
derekbit: ReplicaInstance will be used for legacy and spdk volumes' replica directories or spdk replica. I think a it's more clear for them.

pkg/proxy/proxy.go
```diff
@@ -14,13 +14,18 @@ type Proxy struct {
 	logsDir       string
 	shutdownCh    chan error
 	HealthChecker HealthChecker
+
+	diskServiceAddress string
```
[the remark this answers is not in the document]
derekbit: No. This is for experimental design. Forgot removing it. Thank you for reminder.

pkg/imrpc/instance.proto
```diff
@@ -0,0 +1,95 @@
...
+	rpc VersionGet(google.protobuf.Empty) returns(VersionResponse);
+}
+
+message ProcessSpecific {
+	string binary = 1;
+	repeated string args = 2;
+}
+
+message SpdkSpecific {
+	string address = 1;
```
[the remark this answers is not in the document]
derekbit: Old design. Removed, thanks.

pkg/disk/disk.go
```diff
@@ -0,0 +1,210 @@
...
+func (s *Server) DiskGet(ctx context.Context, req *rpc.DiskGetRequest) (*rpc.Disk, error) {
+	log := logrus.WithFields(logrus.Fields{
+		"diskType": req.DiskType,
+		"diskName": req.DiskName,
+		"diskPath": req.DiskPath,
+	})
+
+	log.Info("Getting disk info")
+
+	if req.DiskName == "" || req.DiskPath == "" {
```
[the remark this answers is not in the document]
derekbit: Just used for double check.

pkg/imrpc/imcommon.proto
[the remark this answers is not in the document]
shuo-wu: Yeah there is no much gain... I was thinking it may reduce one entry for most of the requests. Let's ignore it now

pkg/client/instance.go
```diff
@@ -0,0 +1,311 @@
...
+	if !ok {
+		return nil, fmt.Errorf("failed to delete instance: invalid backend store driver %v", req.BackendStoreDriver)
+	}
+
+	client := c.getControllerServiceClient()
+	ctx, cancel := context.WithTimeout(context.Background(), types.GRPCServiceTimeout)
+	defer cancel()
+
+	p, err := client.InstanceCreate(ctx, &rpc.InstanceCreateRequest{
+		Spec: &rpc.InstanceSpec{
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

pkg/client/instance.go
```diff
@@ -0,0 +1,311 @@
...
+
+	client := c.getControllerServiceClient()
+	ctx, cancel := context.WithTimeout(context.Background(), types.GRPCServiceTimeout)
+	defer cancel()
+
+	p, err := client.InstanceGet(ctx, &rpc.InstanceGetRequest{
+		Name:               name,
+		Type:               instanceType,
+		BackendStoreDriver: rpc.BackendStoreDriver(driver),
+		DiskUuid:           diskUUID,
```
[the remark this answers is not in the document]
derekbit: Removed. Thank you.

pkg/disk/disk.go
```diff
@@ -0,0 +1,210 @@
...
+	return &rpc.VersionResponse{
+		Version:   v.Version,
+		GitCommit: v.GitCommit,
+		BuildDate: v.BuildDate,
+
+		InstanceManagerAPIVersion:    int64(v.InstanceManagerAPIVersion),
+		InstanceManagerAPIMinVersion: int64(v.InstanceManagerAPIMinVersion),
+
+		InstanceManagerProxyAPIVersion:    int64(v.InstanceManagerProxyAPIVersion),
+		InstanceManagerProxyAPIMinVersion: int64(v.InstanceManagerProxyAPIMinVersion),
```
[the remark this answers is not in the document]
derekbit: No, removed. Thank you.

pkg/api/types.go
```diff
@@ -1,9 +1,82 @@
...
+		Type:               obj.Spec.Type,
+		BackendStoreDriver: obj.Spec.BackendStoreDriver.String(),
+		PortCount:          obj.Spec.PortCount,
+		PortArgs:           obj.Spec.PortArgs,
+		InstanceStatus:     RPCToInstanceStatus(obj.Status),
+	}
+
+	if obj.Spec.ProcessSpecific != nil {
+		instance.Binary = obj.Spec.ProcessSpecific.Binary
+		instance.Args = obj.Spec.ProcessSpecific.Args
```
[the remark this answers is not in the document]
derekbit: I think we don't need to make `type Instance struct` complicated. 
`Args` is not used,  and `Binary` is only used by `InstanceGetBinary`.

```
type Instance struct {
	Name               string   `json:"name"`
	Type               string   `json:"type"`
	BackendStoreDriver string   `json:"backendStoreDriver"`
	Binary             string   `json:"binary"`
	Args               []string `json:"args"`
	PortCount          int32    `json:"portCount"`
	PortArgs           []string `json:"portArgs"`

	InstanceStatus InstanceStatus `json:"instanceStatus"`

	Deleted bool `json:"deleted"`
}
```

Dockerfile.dapper
```diff
@@ -44,22 +40,6 @@ ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm64=arm64 GOLANG_ARCH_s390x=s390x GOLA
...
-    make; \
-    make install
-
-# Build TGT
-RUN cd /usr/src && \
-    git clone https://github.com/rancher/tgt.git && \
-    cd tgt && \
-    git checkout 3a8bc4823b5390e046f7aa8231ed262c0365c42c && \
-    make; \
-    make install
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

pkg/api/types.go
```diff
@@ -1,9 +1,82 @@
 package api
 
 import (
+	"github.com/golang/protobuf/ptypes/empty"
+	spdkrpc "github.com/longhorn/longhorn-spdk-engine/proto/spdkrpc"
+
 	rpc "github.com/longhorn/longhorn-instance-manager/pkg/imrpc"
 )
 
+type Instance struct {
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

pkg/disk/disk.go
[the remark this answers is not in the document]
shuo-wu: Then I prefer to remove it if the input is just for verification. IIUC, the upper disk server will do verification periodically

pkg/api/types.go
shuo-wu: I see. This is unrelated to process/spdk instances. It's actually the entries of the disk ReplicaMap. I am thinking if it's better to rename it.

pkg/api/types.go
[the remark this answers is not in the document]
shuo-wu: OK. I prefer to use separate `struct` for spdk and process. This definition looks messy...
WDYT? @innobead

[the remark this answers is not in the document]
derekbit: Using separate structs might beome more messy. `Instance` means `process` or `spdk instance`.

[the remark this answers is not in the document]
derekbit: > OK. I prefer to use separate `struct` for spdk and process. This definition looks messy... WDYT? @innobead

OK, we can separate them.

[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

pkg/api/types.go
derekbit: Rename to `ReplicaStorageInstance`

pkg/disk/disk.go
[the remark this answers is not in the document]
derekbit: Removed, thanks.

pkg/instance/instance.go
derekbit: Yeah, we still need TLS for ProcessManagerClient because of the backward compatibility.

pkg/imrpc/imcommon.proto
```diff
@@ -0,0 +1,30 @@
...
+
+    int64 instanceManagerDiskServiceAPIVersion = 8;
+    int64 instanceManagerDiskServiceAPIMinVersion = 9;
+
+    int64 instanceManagerInstanceServiceAPIVersion = 10;
+    int64 instanceManagerInstanceServiceAPIMinVersion =11;
+}
+
+message LogResponse {
+	string line = 2;
```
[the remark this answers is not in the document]
derekbit: The check is blocked by https://github.com/longhorn/longhorn/issues/6012 and not able to check the compatibility...
Keep it as is for now.

pkg/instance/instance.go
```diff
@@ -0,0 +1,588 @@
...
+		}
+	}
+	return nil
+}
+
+func (s *Server) spdkInstanceLog(req *rpc.InstanceLogRequest, srv rpc.InstanceService_InstanceLogServer) error {
+	return fmt.Errorf("spdk instance log is not supported")
+}
+
+func (s *Server) InstanceWatch(req *empty.Empty, srv rpc.InstanceService_InstanceWatchServer) error {
```
shuo-wu, line 413: Does this instance server needs to implement the broadcasting mechanism?

pkg/disk/disk.go
```diff
@@ -0,0 +1,271 @@
...
+	})
+
+	log.Info("Creating disk")
+
+	if req.DiskName == "" || req.DiskPath == "" {
+		return nil, grpcstatus.Error(grpccodes.InvalidArgument, "disk name and disk path are required")
+	}
+
+	s.Lock()
+	defer s.Unlock()
```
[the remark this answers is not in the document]
derekbit: Yes, should be removed. This is the old design.

pkg/instance/instance.go
derekbit: Not necessary. Any concern?

pkg/disk/disk.go
```diff
@@ -0,0 +1,271 @@
...
+		return spdkDiskToDisk(ret), nil
+	case rpc.DiskType_filesystem:
+		// TODO: implement filesystem disk type
+		fallthrough
+	default:
+		return nil, grpcstatus.Errorf(grpccodes.Unimplemented, "unsupported disk type %v", req.DiskType)
+	}
+}
+
+func (s *Server) DiskDelete(ctx context.Context, req *rpc.DiskDeleteRequest) (*emptypb.Empty, error) {
```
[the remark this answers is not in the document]
derekbit: Cannot provide the`req.DiskType`. `node.diskStatus` doesn't contain the `diskType`, so `DiskDelete` directly sends diskName and diskUUID to spdk engine.

pkg/disk/disk.go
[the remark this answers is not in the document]
derekbit: Yes, should be removed. This is the old design for the race condition in previous spdk engine.

pkg/instance/instance.go
```diff
@@ -0,0 +1,588 @@
...
+}
+
+func (s *Server) watchSPDKReplica(ctx context.Context, req *emptypb.Empty, srv rpc.InstanceService_InstanceWatchServer) error {
+	logrus.Info("Start watching SPDK replica")
+
+	c, err := spdkclient.NewSPDKClient(s.spdkServiceAddress)
+	if err != nil {
+		return errors.Wrap(err, "failed to create DiskServiceClient")
+	}
+	defer c.Close()
```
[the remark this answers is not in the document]
derekbit: Fixed. Thank you.

pkg/api/instance.go
```diff
@@ -0,0 +1,128 @@
+package api
+
+import (
+	"github.com/golang/protobuf/ptypes/empty"
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

pkg/disk/disk.go
```diff
@@ -0,0 +1,271 @@
...
+	log.Info("Deleting disk")
+
+	if req.DiskName == "" || req.DiskUuid == "" {
+		return &empty.Empty{}, grpcstatus.Error(grpccodes.InvalidArgument, "disk name and disk UUID are required")
+	}
+
+	s.Lock()
+	defer s.Unlock()
+
+	if s.spdkEnabled {
```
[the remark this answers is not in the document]
derekbit: The check can be removed.

pkg/instance/instance.go
shuo-wu: I see. This instance watch implementation blindly starts new connections with the below process & spdk servers then forwards the messages to the caller.
```
monitor thread A in longhorn manager == Instance Server == Process/SPDK Server
monitor thread B in longhorn manager == Instance Server == Process/SPDK Server
```

The ideal implementation is like this, which always uses a single connection between Instance Server and Process/SPDK server to handle all watch streaming:
```
monitor thread A in longhorn manager  
                                                                 \
                                                                   == Instance Server == Process/SPDK Server
                                                                 /
monitor thread B in longhorn manager 
```
But it's a little bit complicated while the gain is little. We can ignore it for now.

pkg/disk/disk.go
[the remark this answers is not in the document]
shuo-wu: Considering we will move the legacy disk handling to this server, should we record `diskType` in `node.diskStatus` and input it to this API?

pkg/disk/disk.go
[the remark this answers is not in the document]
shuo-wu: Then let's remove it

pkg/disk/disk.go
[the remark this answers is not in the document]
derekbit: Yeah, if we don't care about the duplicate in spec and status. The solution is simplest. Let's do it this way, ok?

pkg/api/process.go
```diff
@@ -61,28 +61,6 @@ func NewProcessStream(stream rpc.ProcessManagerService_ProcessWatchClient) *Proc
...
-	stream rpc.ProcessManagerService_ProcessLogClient
-}
-
-func (s *LogStream) Recv() (string, error) {
-	resp, err := s.stream.Recv()
-	if err != nil {
-		return "", err
-	}
-	return resp.Line, nil
+func (s *ProcessStream) Recv() (*rpc.ProcessResponse, error) {
```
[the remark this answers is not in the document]
derekbit: Sure. Thank you.

pkg/client/instance.go
```diff
@@ -0,0 +1,318 @@
...
+	client := c.getControllerServiceClient()
+	stream, err := client.InstanceWatch(ctx, &empty.Empty{})
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to open instance update stream")
+	}
+
+	return api.NewInstanceStream(stream), nil
+}
+
+func (c *InstanceServiceClient) InstanceReplace(backendStoreDriver, name, instanceType, binary string, portCount int, args, portArgs []string, terminateSignal string) (*api.Instance, error) {
```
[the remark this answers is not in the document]
derekbit: Yeah, the server returns an error
```
func (s *Server) spdkInstanceReplace(req *rpc.InstanceReplaceRequest) (*rpc.InstanceResponse, error) {
	return nil, fmt.Errorf("spdk instance replace is not supported")
}
```

pkg/proxy/backup.go
```diff
@@ -73,10 +90,29 @@ func (p *Proxy) SnapshotBackup(ctx context.Context, req *rpc.EngineSnapshotBacku
...
 	log.Tracef("Getting %v backup status from replica %v", req.BackupName, req.ReplicaAddress)
 
+	switch req.ProxyEngineRequest.BackendStoreDriver {
+	case rpc.BackendStoreDriver_longhorn:
+		return p.snapshotBackupStatus(ctx, req)
+	case rpc.BackendStoreDriver_spdk:
+		return p.spdkSnapshotBackupStatus(ctx, req)
+	default:
+		return nil, grpcstatus.Errorf(grpccodes.InvalidArgument, "unknown backend store driver %v", req.ProxyEngineRequest.BackendStoreDriver)
+	}
```
innobead, line 112: nit: this could be improved by reflection to reduce the repeated code. Can be improved later.

pkg/proxy/proxy.go
```diff
@@ -35,7 +40,6 @@ func (p *Proxy) startMonitoring() {
 		case <-p.shutdownCh:
 			logrus.Info("Proxy Server is shutting down")
 			done = true
```
innobead, line 42: nit: var `done` is not necessary.

pkg/imrpc/instance.proto
```diff
@@ -0,0 +1,92 @@
+syntax="proto3";
+
+package imrpc;
+
+import "google/protobuf/empty.proto";
+import "imcommon.proto";
+
+service InstanceService {
+    rpc InstanceCreate(InstanceCreateRequest) returns (InstanceResponse) {}
+	rpc InstanceDelete(InstanceDeleteRequest) returns (InstanceResponse) {}
```
innobead, line 10: Make sure the code formatted. It seems wrong indent here.

pkg/imrpc/instance.proto
```diff
@@ -0,0 +1,92 @@
...
+	rpc InstanceGet(InstanceGetRequest) returns (InstanceResponse) {}
+	rpc InstanceList(google.protobuf.Empty) returns (InstanceListResponse) {}
+	rpc InstanceLog(InstanceLogRequest) returns (stream LogResponse) {}
+	rpc InstanceWatch(google.protobuf.Empty) returns (stream google.protobuf.Empty) {}
+	rpc InstanceReplace(InstanceReplaceRequest) returns (InstanceResponse) {}
+
+	rpc VersionGet(google.protobuf.Empty) returns(VersionResponse);
+}
+
+message ProcessSpecific {
```
innobead, line 20: `ProcessSpecific` seems the word is not finished. How about `ProcessInstace` or `ProcessInstanceSpec`?

pkg/imrpc/instance.proto
```diff
@@ -0,0 +1,92 @@
...
+	string binary = 1;
+	repeated string args = 2;
+}
+
+message SpdkSpecific {
```
innobead, line 25: `SpdkInstance` or `SpdkInstanceSpec`?

pkg/instance/healthchecker.go
```diff
@@ -0,0 +1,43 @@
...
+		case <-stopCh:
+			logrus.Infof("Stop waiting for gRPC service of instance service %v to start at %v", name, address)
+			return false
+
+		case <-ticker.C:
+			if c.IsRunning(address) {
+				logrus.Infof("Instance service %v has started at %v", name, address)
+				return true
+			}
+			logrus.Infof("Wait for gRPC service of instance service %v to start at %v", name, address)
```
innobead, line 38: `Waiting for ....`

pkg/instance/healthchecker.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func (c *GRPCHealthChecker) WaitForRunning(address, name string, stopCh chan struct{}) bool {
+	ticker := time.NewTicker(types.WaitInterval)
+	defer ticker.Stop()
+
+	for i := 0; i < types.WaitCount; i++ {
+		select {
+		case <-stopCh:
+			logrus.Infof("Stop waiting for gRPC service of instance service %v to start at %v", name, address)
```
innobead, line 30: `Stopped waiting ...`

pkg/instance/instance.go
```diff
@@ -0,0 +1,588 @@
...
+	return s, nil
+}
+
+func (s *Server) startMonitoring() {
+	done := false
+	for {
+		select {
+		case <-s.shutdownCh:
+			logrus.Info("Instance Server is shutting down")
+			done = true
```
innobead, line 57: nit: done is not necessary, can just use break.

pkg/imrpc/instance.proto
```diff
@@ -0,0 +1,92 @@
...
+package imrpc;
+
+import "google/protobuf/empty.proto";
+import "imcommon.proto";
+
+service InstanceService {
+    rpc InstanceCreate(InstanceCreateRequest) returns (InstanceResponse) {}
+	rpc InstanceDelete(InstanceDeleteRequest) returns (InstanceResponse) {}
+	rpc InstanceGet(InstanceGetRequest) returns (InstanceResponse) {}
+	rpc InstanceList(google.protobuf.Empty) returns (InstanceListResponse) {}
```
innobead, line 12: nit: ideally we should use the common naming like `ListInstance`; however, I know we juts follow the existing naming used in the current code base, but I would suggest using the common way instead.

pkg/instance/instance.go
```diff
@@ -0,0 +1,588 @@
...
+		req.Spec.ProcessSpecific.Binary, int(req.Spec.PortCount), req.Spec.ProcessSpecific.Args, req.Spec.PortArgs, req.TerminateSignal)
+	if err != nil {
+		return nil, err
+	}
+
+	return processResponseToInstanceResponse(process), nil
+}
+
+func (s *Server) spdkInstanceReplace(req *rpc.InstanceReplaceRequest) (*rpc.InstanceResponse, error) {
+	return nil, fmt.Errorf("spdk instance replace is not supported")
```
innobead, line 362: Use gRPC error instead. Please review all returned errors in gRPC methods.

pkg/instance/instance.go
```diff
@@ -0,0 +1,588 @@
...
+	}
+
+	if err := g.Wait(); err != nil {
+		return errors.Wrap(err, "failed to watch instances")
+	}
+
+	return nil
+}
+
+func (s *Server) watchSPDKReplica(ctx context.Context, req *emptypb.Empty, srv rpc.InstanceService_InstanceWatchServer) error {
```
innobead, line 440: `Starting watching...`

pkg/instance/instance.go
```diff
@@ -0,0 +1,588 @@
...
+	logrus.Info("Start watching SPDK replica")
+
+	c, err := spdkclient.NewSPDKClient(s.spdkServiceAddress)
+	if err != nil {
+		return errors.Wrap(err, "failed to create SPDKServiceClient")
```
innobead, line 445: return gRPC error instead.

app/cmd/start.go
```diff
@@ -112,94 +130,234 @@ func start(c *cli.Context) (err error) {
...
+		return "", "", "", "", "", err
 	}
 
 	intPort, err := strconv.Atoi(port)
 	if err != nil {
-		return "", err
+		return "", "", "", "", "", err
+	}
+
+	return net.JoinHostPort(host, strconv.Itoa(intPort)),
```
innobead, line 234: Just make the port should be compatible with old IM like process manager and proxy.

pkg/util/util.go
```diff
@@ -100,3 +101,22 @@ func Now() string {
 func UUID() string {
 	return uuid.New().String()
 }
+
+func ParsePortRange(portRange string) (int32, int32, error) {
```
innobead, line 107: Suggest adding unit test.

generate_grpc.sh
```diff
@@ -18,10 +18,23 @@ if [ ! -e ${DIR} ]; then
...
 python3 -m grpc_tools.protoc -I pkg/imrpc -I proto/vendor/protobuf/src/ --python_out=integration/rpc/instance_manager --grpc_python_out=integration/rpc/instance_manager pkg/imrpc/imrpc.proto
 protoc -I pkg/imrpc/ -I proto/vendor/protobuf/src/ pkg/imrpc/imrpc.proto --go_out=plugins=grpc:pkg/imrpc
 
 # proxy
 python3 -m grpc_tools.protoc -I pkg/imrpc -I proto/vendor/ -I proto/vendor/protobuf/src/ --python_out=integration/rpc/proxy --grpc_python_out=integration/rpc/proxy pkg/imrpc/proxy.proto
 protoc -I pkg/imrpc/ -I proto/vendor/ -I proto/vendor/protobuf/src/ pkg/imrpc/proxy.proto --go_out=plugins=grpc:pkg/imrpc/
+
+# disk
+python3 -m grpc_tools.protoc -I pkg/imrpc -I proto/vendor/ -I proto/vendor/protobuf/src/ --python_out=integration/rpc/disk --grpc_python_out=integration/rpc/disk pkg/imrpc/disk.proto
+protoc -I pkg/imrpc/ -I proto/vendor/ -I proto/vendor/protobuf/src/ pkg/imrpc/disk.proto --go_out=plugins=grpc:pkg/imrpc/
```
innobead, line 35: Let's make sure proto registration is unique. ref: https://github.com/longhorn/backing-image-manager/commit/38ff27cc

innobead (changes requested):
Some feedback first. In general, LGTM, but stlll need to review disk service part.

package/Dockerfile
```diff
@@ -23,6 +22,19 @@ COPY bin/longhorn-instance-manager /usr/local/bin/
 
 COPY package/instance-manager /usr/local/bin/
 
+# Build go-spdk-helper
+ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm64=arm64 GOLANG_ARCH_s390x=s390x GOLANG_ARCH=GOLANG_ARCH_${ARCH} \
```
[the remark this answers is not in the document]
derekbit: It is an existing issue, because development tools such as gcc are installed for compilation and are not removed.
Will improve in the future.

pkg/client/proxy_volume.go
```diff
@@ -124,21 +159,32 @@ func (c *ProxyClient) VolumeFrontendShutdown(serviceAddress string) (err error)
...
 }
 
-func (c *ProxyClient) VolumeUnmapMarkSnapChainRemovedSet(serviceAddress string, enabled bool) (err error) {
+func (c *ProxyClient) VolumeUnmapMarkSnapChainRemovedSet(backendStoreDriver, engineName, serviceAddress string, enabled bool) (err error) {
 	input := map[string]string{
+		"engineName":     engineName,
 		"serviceAddress": serviceAddress,
+		"enabled":        strconv.FormatBool(enabled),
 	}
 	if err := validateProxyMethodParameters(input); err != nil {
```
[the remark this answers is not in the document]
derekbit: I don't know the context indeed...
cc @c3y1huang

pkg/imrpc/instance.proto
derekbit: Weird VSCode... It said it is formatted.

pkg/imrpc/instance.proto
derekbit: `ProcessSpecific` means they are **specific** for `Process`.
`ProcessInstace` or `ProcessInstanceSpec` sounds not suitable.

pkg/imrpc/instance.proto
derekbit: Same reason as https://github.com/longhorn/longhorn-instance-manager/pull/218#discussion_r1209738210

pkg/imrpc/instance.proto
innobead: I see. How about `ProcessSpec`? `ProcessSpecific` seems an adjective, which is not quite suitable for a class name. WDYT?

pkg/disk/disk.go
```diff
@@ -0,0 +1,254 @@
...
+
+	switch req.DiskType {
+	case rpc.DiskType_block:
+		ret, err := s.spdkClient.DiskCreate(req.DiskName, req.DiskPath, req.BlockSize)
+		if err != nil {
+			return nil, grpcstatus.Error(grpccodes.Internal, err.Error())
+		}
+		return spdkDiskToDisk(ret), nil
+	case rpc.DiskType_filesystem:
+		// TODO: implement filesystem disk type
```
innobead, line 114: Let's create a ticket for this in the future.

pkg/disk/disk.go
```diff
@@ -0,0 +1,254 @@
...
+
+	log.Info("Listing disk replica instances")
+
+	if req.DiskName == "" {
+		return nil, grpcstatus.Error(grpccodes.InvalidArgument, "disk name is required")
+	}
+
+	switch req.DiskType {
+	case rpc.DiskType_block:
+		replicas, err := s.spdkClient.ReplicaList()
```
innobead, line 182: Shouldn't the disk name be used here? the code here is just listing all replicas from the spdk engine.

pkg/disk/disk.go
```diff
@@ -0,0 +1,254 @@
...
+
+	log.Info("Deleting disk replica instance")
+
+	if req.DiskName == "" || req.DiskUuid == "" || req.ReplciaInstanceName == "" {
+		return nil, grpcstatus.Error(grpccodes.InvalidArgument, "disk name, disk UUID and replica instance name are required")
+	}
+
+	switch req.DiskType {
+	case rpc.DiskType_block:
+		err := s.spdkClient.ReplicaDelete(req.ReplciaInstanceName, true)
```
innobead, line 217: The same question. Shouldn't disk params be used here for replica deletion? or not really required.

pkg/util/util.go
```diff
@@ -100,3 +103,34 @@ func Now() string {
...
+
+	portEnd, err := strconv.Atoi(strings.TrimSpace(parts[1]))
+	if err != nil {
+		return 0, 0, err
+	}
+
+	return int32(portStart), int32(portEnd), nil
+}
+
+func IsSPDKTgtReady(timeout time.Duration) bool {
```
innobead, line 126: Suggest adding doc comment for any exported functions/methods.

innobead (changes requested):
In general, LGTM. Most of feedbacks are general and some are optional or can be improved later. 

After resolving them, let's merge this PR.

pkg/imrpc/instance.proto
derekbit: I'm fine with it.

pkg/disk/disk.go
```diff
@@ -0,0 +1,254 @@
...
+		return spdkDiskToDisk(ret), nil
+	case rpc.DiskType_filesystem:
+		// TODO: implement filesystem disk type
+		fallthrough
+	default:
+		return nil, grpcstatus.Errorf(grpccodes.Unimplemented, "unsupported disk type %v", req.DiskType)
+	}
+}
+
+func (s *Server) DiskDelete(ctx context.Context, req *rpc.DiskDeleteRequest) (*emptypb.Empty, error) {
```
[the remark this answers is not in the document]
derekbit: DiskStatus doesn't have diskType. I will add it.

pkg/disk/disk.go
derekbit: We have a replicaMap (cache) where replica names serve as keys, so the disk name is not used.

derekbit: We have a replicaMap (cache) where replica names server as keys.
ReplicaList doesn't support filter by diskName yet.

pkg/disk/disk.go
derekbit: We have a replicaMap (cache) where replica names server as keys, so no need to send disk name.

app/cmd/start.go
derekbit: Sure.
8500: ProcessManagerService
8501: ProxyService
8502: DiskService
8503: InstanceService
8504: SpdkService

pkg/imrpc/imcommon.proto
```diff
@@ -0,0 +1,30 @@
+syntax="proto3";
+
+package imrpc;
+
+message VersionResponse {
```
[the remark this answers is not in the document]
derekbit: Same as before.
`instanceManagerAPIVersion` >= 4 will use `instance` gRPC service.

pkg/process/process_manager.go
```diff
@@ -88,7 +88,6 @@ func (pm *Manager) startMonitoring() {
 		case <-pm.shutdownCh:
 			logrus.Info("Process Manager is shutting down")
 			done = true
```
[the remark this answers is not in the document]
derekbit: Fixed, thank you.

pkg/proxy/proxy.go
derekbit: Fixed, thank you.

pkg/imrpc/instance.proto
derekbit: Addressed. Thank you.

pkg/instance/healthchecker.go
derekbit: Addressed. Thank you.

pkg/instance/instance.go
derekbit: Removed. Thank you.

pkg/instance/instance.go
derekbit: Addressed. Thank you.

pkg/instance/instance.go
derekbit: Addressed. Thank you.

generate_grpc.sh
derekbit: Addressed. Thank you.

pkg/disk/disk.go
```diff
@@ -0,0 +1,254 @@
...
+	return s, nil
+}
+
+func (s *Server) startMonitoring() {
+	done := false
+	for {
+		select {
+		case <-s.shutdownCh:
+			logrus.Info("Disk gRPC Server is shutting down")
+			done = true
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

pkg/imrpc/instance.proto
derekbit: Addressed. Thank you.

pkg/instance/healthchecker.go
derekbit: Fixed, thank you.

pkg/instance/instance.go
derekbit: So,
The client codes still need TLS because of the communication to old IMs. TLS support for the process manager server can be removed. Right?

pkg/disk/disk.go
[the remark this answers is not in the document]
derekbit: Removed. Thank you.

pkg/disk/disk.go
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

package/Dockerfile
```diff
@@ -39,6 +38,20 @@ RUN cd /usr/src && \
     make; \
     make install
 
+# Build SPDK
+ENV SPDK_DIR /usr/src/spdk
+RUN if [ ${ARCH} == "amd64" ]; then \
```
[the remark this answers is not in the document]
derekbit: Tracked in https://github.com/longhorn/longhorn/issues/6021

package/Dockerfile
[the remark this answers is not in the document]
derekbit: Tracked in https://github.com/longhorn/longhorn/issues/6022

pkg/disk/disk.go
derekbit: Tracked in https://github.com/longhorn/longhorn/issues/6024

pkg/disk/disk.go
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

derekbit:
Ready for review. I will update longhorn-manager later. Thank you. 
cc @innobead @shuo-wu

innobead (approved):

innobead:
Waiting for longhorn-manager PR ready before merging this. cc @shuo-wu

shuo-wu:
Just re-checked the new comments and some changes. In general both PRs LGTM.

PR #229, opened 2023-05-11
Engine identity validation

ejweber:
longhorn/longhorn#5845

This PR makes use of the changes in longhorn-engine (https://github.com/longhorn/longhorn-engine/pull/902) to ensure identity validation during normal Longhorn operation.


pkg/meta/version.go
```diff
@@ -6,7 +6,7 @@ const (
 	InstanceManagerAPIMinVersion = 1
 
 	// InstanceManagerProxyAPIVersion is used for compatibility check for longhorn-manager
-	InstanceManagerProxyAPIVersion    = 4
```
shuo-wu, line 9: This API version is already bumped for v1.5.0. The issue ticket number is 5376

pkg/client/proxy_snapshot.go
```diff
@@ -120,13 +120,14 @@ func (c *ProxyClient) SnapshotList(backendStoreDriver, engineName, volumeName,
...
-		"volumeName":     volumeName,
-		"serviceAddress": serviceAddress,
-		"snapshotName":   snapshotName,
-		"fromController": fromController,
+		"engineName":            engineName,
+		"volumeName":            volumeName,
+		"serviceAddress":        serviceAddress,
+		"snapshotName":          snapshotName,
+		"fromControllerAddress": fromControllerAddress,
+		"fromControllerName":    fromControllerName,
```
shuo-wu: NIT: `fromControllerName` vs `engineName` looks weird... Can you help correct this part with uniform naming?

pkg/imrpc/proxy.proto
```diff
@@ -164,6 +164,7 @@ message EngineSnapshotBackupStatusRequest {
 
 	string backup_name = 2;
 	string replica_address = 3;
+	string replica_name = 4;
```
shuo-wu, line 167: Not sure if it's better to add this to `ControllerReplica`. So far the SPDK engine/controller will tell the replica address along with the name.

shuo-wu (dismissed):
In general LGTM. This PR mainly adds `volume_name` and `engine_name`/`replica_name` for instance proxy APIs

pkg/meta/version.go
ejweber: I do not anticipate this change being added to v1.5.0, so I am thinking an update is still required.

pkg/imrpc/proxy.proto
ejweber: In general, I think it WOULD be better. However, as you are alluding to, it requires us to refactor the in-memory data structures the longhorn-engine controller server uses to keep track of replicas (currently, the v1 engine itself has no concept of the name of a replica). I am glad the v2 engine is enhanced to understand replica name in its initial release, as this will prevent a number of problems. However, the v1 refactoring required for this suggestion is out-of-scope in the existing LEP: https://github.com/longhorn/longhorn/blob/master/enhancements/20230420-engine-identity-validation.md#non-goals.

pkg/meta/version.go
shuo-wu: Oh OK. Then bumping up the version is required. BTW, should we increase `CLIAPIVersion` number either?

ejweber: We handle this in longhorn-engine, unless there is somewhere else it is also necessary?

https://github.com/longhorn/longhorn-engine/blob/3af37b9e5bf984204c440fc9aa6ce9badbd0c8ad/pkg/meta/version.go#L5

pkg/imrpc/proxy.proto
```diff
@@ -119,10 +120,11 @@ message EngineSnapshotPurgeStatusProxyResponse {
 message EngineSnapshotCloneRequest {
 	ProxyEngineRequest proxy_engine_request = 1;
 
-	string from_controller = 2;
```
shuo-wu, line 122: IIUC, dIrectly changing this input means that we need to bump up `InstanceManagerProxyAPIMinVersion` as well.

ejweber: I don't think so. We are not changing the way the field is used at all, nor are we changing its tag. Clients should put the address of the snapshot's source engine on the wire with tag 2 and servers should pull it off the wire with tag 2 regardless of how it is named. (https://earthly.dev/blog/backward-and-forward-compatibility/#adding-new-fields)

However, I only made this change to try to follow through on the request in https://github.com/longhorn/longhorn-instance-manager/pull/229#discussion_r1247852573. I can revert to using `from_controller` if preferred. (IMO this would be better than bumping `InstanceManagerProxyAPIMinVersion`).

ejweber:
@shuo-wu please re-review and @PhanLe1010 please review when you are able.

PhanLe1010 (dismissed):
LGTM

pkg/imrpc/proxy.proto
PhanLe1010: Ah. The beauty of protobuf saves us

ejweber:
There is a new commit that fixes cloning, which is not included in the core tests. My code mistakenly assumed that the volume name we cloned from and to would be the same. This caused us to try to communicate with a from replica using the to volume name (triggering an invalid gRPC error). Please rereview when you are able.

cc @PhanLe1010 @shuo-wu @james-munson

PhanLe1010 (approved):
LGTM

(reviewed the newly added snapshot clone commit)

shuo-wu (dismissed):
LGTM. You can merge it after resolving the conflict.

PhanLe1010 (approved):

PR #245, opened 2023-06-13
Support offline replica rebuilding for SPDK volumes

derekbit:
Longhorn/Longhorn#6067
Longhorn/Longhorn#6071


pkg/proxy/snapshot.go
```diff
@@ -281,7 +281,8 @@ func (p *Proxy) snapshotPurge(ctx context.Context, req *rpc.EngineSnapshotPurgeR
 }
 
 func (p *Proxy) spdkSnapshotPurge(ctx context.Context, req *rpc.EngineSnapshotPurgeRequest) (resp *empty.Empty, err error) {
-	return nil, grpcstatus.Errorf(grpccodes.Unimplemented, "not implemented")
+	/* TODO: implement this */
+	return &empty.Empty{}, nil
```
innobead, line 285: Even though we have the initial implementation in SPDK server, but we have not integrated to here yet. Suggest keeping the original uimpl error first.

innobead (changes requested):
Just one comemnt.

pkg/proxy/snapshot.go
innobead: After discussing with @derekbit , this is for snapshot purge instead I misread it. Basically, this is intended to ignore purging since we will not do anything currently for SPDK volume.

innobead (approved):

innobead:
@mergify backport v1.5.x

PR #258, opened 2023-06-20
Update SPDK repos and improve logs

shuo-wu:

PR #260, opened 2023-06-20
Update SPDK repos and improve logs (backport #258)

shuo-wu:
backport of pull request https://github.com/longhorn/longhorn-instance-manager/pull/258

### DIFF
--- a/pkg/disk/disk.go
+++ b/pkg/disk/disk.go
@@ -45,7 +45,7 @@ func NewServer(spdkEnabled bool, spdkServiceAddress string, shutdownCh chan erro
 	}
 
 	if spdkEnabled {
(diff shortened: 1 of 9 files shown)


pkg/disk/disk.go
```diff
@@ -93,7 +93,7 @@ func (s *Server) DiskCreate(ctx context.Context, req *rpc.DiskCreateRequest) (*r
 		"blockSize": req.BlockSize,
 	})
 
-	log.Info("Creating disk")
+	log.Info("Disk Server: Creating disk")
```
innobead, line 96: nit: can add log owner (disk server) to the log instance instead.

innobead (approved):

innobead:
@shuo-wu also need to update longhorn manager dependencies.

PR #274, opened 2023-08-03
nit: fix incorrect error log

PhanLe1010:


PhanLe1010:
@mergify backport v1.5.x v1.4.x v1.3.x

pkg/process/process.go
```diff
@@ -174,7 +174,7 @@ func (p *Process) StopWithSignal(signal syscall.Signal) {
 			logrus.Infof("Wait for process %v to shutdown", p.Name)
 			time.Sleep(types.WaitInterval)
 		}
-		logrus.Warnf("Process Manager: cannot graceful stop process %v in %v seconds, will kill the process", p.Name, types.WaitCount)
+		logrus.Warnf("Process Manager: cannot graceful stop process %v in %v milliseconds, will kill the process", p.Name, int64(types.WaitCount)*int64(types.WaitInterval))
```
ejweber, line 177: Something looked off to me about this math, so I ran a couple of calculations in Go Playground. In short, I think we are actually calculating a value in nanoseconds here and then putting a milliseconds label on it.

IMO it would probably be better to give the result in seconds (or let the autoformatting of time.Duration do it), but I don't particularly care how its formatted as long as its correct.

Please take a look and tell me what you think.

https://go.dev/play/p/NVBnMHhPwCt

ejweber (changes requested):

pkg/process/process.go
PhanLe1010: Thank you very much. Fixed!

ejweber (approved):
LGTM

PR #275, opened 2023-08-04
Update go-iscsi-helper to version also used in longhorn-manager to incorporate PR#63 containerd fix

andrewd-zededa:
This PR is related to the below go-iscsi-helper PR and issues below:

https://github.com/longhorn/go-iscsi-helper/pull/63/files/8719ab241a45576a2644024b80210b22fe916c05

https://github.com/longhorn/longhorn/issues/5693

When trying latest releases of longhorn (1.4.2, and 1.5.0) the above issue exists and it seems to be due to older versions of go-iscsi-helper included in some repos still.  I took a look across the longhorn repos and noticed that various versions are being included but the version "v0.0.0-20230529082528-4c3270590712" used in longhorn-manager seems to be the newest used and it includes the above PR so i've used that version in my commit here.  

Please let me know if there are any issues with my pr like if I should just update to the latest go-iscsi-helper instead.

Thanks

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,7 +39,7 @@ require (
 	github.com/jmespath/go-jmespath v0.3.0 // indirect
 	github.com/kr/pretty v0.3.1 // indirect
 	github.com/kr/text v0.2.0 // indirect
-	github.com/longhorn/go-iscsi-helper v0.0.0-20230215054929-acb305e1031b // indirect
+	github.com/longhorn/go-iscsi-helper v0.0.0-20230529082528-4c3270590712 // indirect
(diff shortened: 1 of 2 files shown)


shuo-wu:
Would you update vendor `/longhornlonghorn-engine`? [This PR](https://github.com/longhorn/longhorn-engine/pull/920) is already merged.

andrewd-zededa:
@shuo-wu should I update longhorn/longhorn-engine in a separate pr?

shuo-wu:
You can directly update vendor `longhorn-engine`. Then `go-iscsi-helper` should be updated automatically as a dependency.

andrewd-zededa:
Sure, I'll fix it

PR #284, opened 2023-09-14
Error out when continuously receiving errors while watching instances

derekbit:
During instance watching, errors can happen due to gRPC server and client errors or keepalive errors. Whenever the error count reaches a maximum, error out the instance watch function.

Longhorn/longhorn#6578

### DIFF
--- a/pkg/instance/instance.go
+++ b/pkg/instance/instance.go
@@ -2,7 +2,9 @@ package instance
 
 import (
 	"context"
+	"fmt"
 	"io"
+	"time"
 
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/pkg/errors"
@@ -19,6 +21,11 @@ import (
 	rpc "github.com/longhorn/longhorn-instance-manager/pkg/imrpc"
 	"github.com/longhorn/longhorn-instance-manager/pkg/meta"
 	"github.com/longhorn/longhorn-instance-manager/pkg/types"
+)
+
+const (
+	maxMonitorRetryCount     = 10
+	monitorRetryPollInterval = 1 * time.Second
 )
 
 type Server struct {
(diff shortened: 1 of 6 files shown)


pkg/instance/instance.go
```diff
@@ -447,18 +455,40 @@ func (s *Server) watchSPDKReplica(ctx context.Context, req *emptypb.Empty, srv r
...
+			}()
+
+			select {
+			case <-ctx.Done():
+				// If context is done before notifier.Recv() completes, return immediately
+				logrus.Info("Stopped watching SPDK replicas")
+				return ctx.Err()
+			case err := <-notifyRecvChan:
+				if err != nil {
+					logrus.WithError(err).Error("Failed to receive next item in SPDK replica watch")
```
ChanYiLin, line 485: suggest use `Warn` here and below since it is still acceptable before `failureCount` is larger than  `maxMonitorRetryCount`
We already `Error` the message above when `failureCount > maxMonitorRetryCount`

BTW, does the failure occur randomly, or will it continue to happen if it occurs once ?

ChanYiLin:
Would like to know will the instance-manager restart when we error out from `watchProcess`?
What will happen if it `failed to watch instances` ?
Thanks

pkg/instance/instance.go
derekbit: > BTW, does the failure occur randomly, or will it continue to happen if it occurs once ?

Both are possible.

derekbit:
> Would like to know will the instance-manager restart when we error out from `watchProcess`? What will happen if it `failed to watch instances` ? Thanks

Yes, will restart a new watch.

pkg/instance/instance.go
```diff
@@ -447,18 +455,40 @@ func (s *Server) watchSPDKReplica(ctx context.Context, req *emptypb.Empty, srv r
...
-				logrus.WithError(err).Error("Failed to receive next item in SPDK replica watch")
-			} else {
-				if err := srv.Send(&empty.Empty{}); err != nil {
-					return errors.Wrap(err, "failed to send instance response")
+			// Use a channel to receive the result of notifier.Recv()
+			notifyRecvChan := make(chan error, 1)
+
+			go func() {
+				_, err := notifier.Recv()
+				notifyRecvChan <- err
```
shuo-wu, line 475: NIT: close the channel after this sending. 
Same below.

shuo-wu (commented):
I just realized that in function `InstanceWatch` there are 3 goroutinue using the same `srv` to send signals, which is not safe based on the grpc server stream description.

The problematic implementation:
```
g.Go(func() error {
		return s.watchProcess(ctx, req, srv)
	})

	if s.spdkEnabled {
		g.Go(func() error {
			return s.watchSPDKReplica(ctx, req, srv)
		})

		g.Go(func() error {
			return s.watchSPDKEngine(ctx, req, srv)
		})
	}
```

The `ServerStream` interface description in `google.golang.org/grpc/stream.go`:
```
	// It is safe to have a goroutine calling SendMsg and another goroutine
	// calling RecvMsg on the same stream at the same time, but it is not safe
	// to call SendMsg on the same stream in different goroutines.
```

derekbit:
> I just realized that in function InstanceWatch there are 3 goroutinue using the same srv to send signals, which is not safe based on the grpc server stream description.

Do you think if we can use a lock protecting `srv` or have a better implementation?

pkg/instance/instance.go
derekbit: Addressed. Thank you.

pkg/instance/instance.go
```diff
@@ -447,25 +478,48 @@ func (s *Server) watchSPDKReplica(ctx context.Context, req *emptypb.Empty, srv r
...
+				_, err := notifier.Recv()
+				errChan <- err
+				close(errChan)
+			}()
+
+			select {
+			case <-ctx.Done():
+				// If context is done before notifier.Recv() completes, return immediately
+				logrus.Info("Stopped watching SPDK replicas")
+				return ctx.Err()
```
shuo-wu, line 506: This `select` looks duplicated...

shuo-wu: Actually, I am not sure why you prefer this kind of implementation. Are you worried that the potential blocking caused by `notifier.Recv()` makes `<-ctx.Done()` not be handled in time?

derekbit: You're right.
I'm thinking which one of `go func() {...}` and `ticker method you implemented in go-spdk-helper` is better?

derekbit: Using another solution, please help check it. Thank you.

pkg/instance/instance.go
```diff
@@ -406,122 +413,187 @@ func (s *Server) spdkInstanceLog(req *rpc.InstanceLogRequest, srv rpc.InstanceSe
...
 	g.Go(func() error {
-		return s.watchProcess(ctx, req, srv)
+		defer func() {
+			// Close the clients for closing streams and unblocking notifier Recv() with error.
+			pmClient.Close()
+			if s.spdkEnabled {
+				spdkClient.Close()
+			}
+		}()
+		return s.handleNotify(ctx, notifyChan, srv)
```
shuo-wu, line 466: I got what you mean. There are 2 minor issues:
1. This implementation avoids `notifier.Recv()` stuck but it cannot prevent `srv.Send` from blocking...
2. Closing clients twice looks a little bit ugly. Can we make this an unregistration goroutine that blindly closes/unregisters all entries of a sync.list? Then we can add the client to the list as long as we launch a new one.

pkg/instance/instance.go
```diff
@@ -406,122 +413,187 @@ func (s *Server) spdkInstanceLog(req *rpc.InstanceLogRequest, srv rpc.InstanceSe
...
+	if s.spdkEnabled {
+		// Create a client for watching SPDK engines and replicas
+		spdkClient, err = spdkclient.NewSPDKClient(s.spdkServiceAddress)
+		if err != nil {
+			return grpcstatus.Error(grpccodes.Internal, errors.Wrapf(err, "failed to create SPDK client").Error())
+		}
+		defer spdkClient.Close()
+	}
+
+	notifyChan := make(chan struct{})
```
shuo-wu, line 452: NIT: Do we need to make this a buffered one?

pkg/instance/instance.go
derekbit: > This implementation avoids notifier.Recv() stuck but it cannot prevent srv.Send from blocking...

Any method for this..? It seems a common issue in a stream server?

pkg/instance/instance.go
derekbit: Yeah, buffered channel sounds good. Use `make(chan struct{}, 1024)` instead.

pkg/instance/instance.go
derekbit: For 2., use a go routine for closing the clients.

pkg/instance/instance.go
```diff
@@ -406,122 +413,205 @@ func (s *Server) spdkInstanceLog(req *rpc.InstanceLogRequest, srv rpc.InstanceSe
...
+		}
+		close(done)
+	}()
+
+	// Create a client for watching processes
+	pmClient, err := client.NewProcessManagerClient("tcp://"+s.processManagerServiceAddress, nil)
+	if err != nil {
+		done <- struct{}{}
+		return grpcstatus.Error(grpccodes.Internal, errors.Wrapf(err, "failed to create ProcessManagerClient").Error())
+	}
```
innobead, line 459: The error message doesn't include the context of the instance. Just noticed most of the logs here have no context, so it will cause logs of each instance monitor same and are unable to differentiate. Let's create a ticket for later improvement.

pkg/instance/instance.go
```diff
@@ -406,122 +413,205 @@ func (s *Server) spdkInstanceLog(req *rpc.InstanceLogRequest, srv rpc.InstanceSe
...
+		logrus.Info("Stopped clients")
+		for name, c := range clients {
+			switch c := c.(type) {
+			case *client.ProcessManagerClient:
+				c.Close()
+			case *spdkclient.SPDKClient:
+				c.Close()
+			}
+			delete(clients, name)
+		}
```
innobead, line 450: Curious if we need to clean up clients here? I assume it's just a variable in the method stack, so it should be automatically free after being out from `InstanceWatch`?

pkg/instance/instance.go
```diff
@@ -406,122 +413,205 @@ func (s *Server) spdkInstanceLog(req *rpc.InstanceLogRequest, srv rpc.InstanceSe
...
-				}
+				notifyChan <- struct{}{}
 			}
 		}
 	}
 }
 
-func (s *Server) watchProcess(ctx context.Context, req *emptypb.Empty, srv rpc.InstanceService_InstanceWatchServer) error {
+func (s *Server) watchProcess(ctx context.Context, req *emptypb.Empty, client *client.ProcessManagerClient, notifyChan chan struct{}) error {
 	logrus.Info("Start watching processes")
```
innobead, line 584: nit: These watch methods look quite similar. Could consolidate them later.

pkg/instance/instance.go
```diff
@@ -21,6 +23,11 @@ import (
 	"github.com/longhorn/longhorn-instance-manager/pkg/types"
 )
 
+const (
+	maxMonitorRetryCount = 10
+	pollInterval         = 1 * time.Second
+)
```
innobead, line 29: Suggest renaming `pollInterval` to `monitorRetryPollInterval`

innobead (changes requested):
In general, LGTM. Just a few comments.

pkg/instance/instance.go
derekbit: Yes, deleting it from a map is not necessary. Just make sure the clients only stores active clients, even it is not used anymore. WDYT?

innobead: Sounds good, it's always good to ensure that. However, just clarified that everything is created in a stack memory, so they should be freed eventually after being out from it.

derekbit: >  I assume it's just a variable in the method stack, so it should be automatically free after being out from InstanceWatch?

Yeah, golang can automatically free the unused stack (garbage collection).

pkg/instance/instance.go
derekbit: Tracked in https://github.com/longhorn/longhorn/issues/6724

pkg/instance/instance.go
derekbit: Updated. Thank you.

pkg/instance/instance.go
derekbit: NP! Thank you.

innobead (approved):

innobead:
@mergify backport v1.5.x v1.4.x

PR #286, opened 2023-09-15
feat: add snapshot max count and size

FrankYang0529:
ref: https://github.com/longhorn/longhorn/issues/6563

rely on: https://github.com/longhorn/longhorn-engine/pull/933


pkg/proxy/volume.go
```diff
@@ -247,3 +249,85 @@ func (p *Proxy) spdkVolumeUnmapMarkSnapChainRemovedSet(ctx context.Context, req
...
+		return nil, err
+	}
+
+	return &emptypb.Empty{}, nil
+}
+
+func (p *Proxy) spdkVolumeMaximumSnapshotCountSet(ctx context.Context, req *rpc.EngineVolumeMaximumSnapshotCountSetRequest) (resp *emptypb.Empty, err error) {
+	/* Not implemented */
+	return &emptypb.Empty{}, nil
+}
```
innobead, line 292: We don't need space management for v2 volume, since it is using the whole disk instead. That means it should have itself space management inside.

innobead (dismissed):
In general, LGTM. Waiting for engine PR merged first.

innobead (dismissed):
LGTM. Merge after the engine PR is merged and the dep is updated.

innobead (approved):

PR #292, opened 2023-09-19
Upgrade package versions

FrankYang0529:
* protoc: 24.3
* protoc-gen-go: 1.5.3
* google.golang.org/grpc: v1.58.1
* empty.proto 24.3
* replace github.com/golang/protobuf with google.golang.org/protobuf

ref: https://github.com/longhorn/longhorn/issues/6666


proto/vendor/github.com/longhorn/longhorn-engine/proto/ptypes/common.proto
```diff
@@ -2,8 +2,10 @@ syntax="proto3";
 
 package ptypes;
 
+option go_package = "github.com/longhorn/longhorn-engine/proto/ptypes";
```
innobead, line 5: I assume we need to update longhorn-engine repo afterward.

pkg/imrpc/common.proto
```diff
@@ -2,6 +2,8 @@ syntax="proto3";
 
 package imrpc;
 
+option go_package = "./;imrpc";
```
innobead, line 5: Why not just use the canonical import path instead? This seems a relative path, so how it will be generated with the right path?

innobead (changes requested):
In general, LGTM. Some comments need to be answered and resolved first.

pkg/imrpc/proxy.pb.go
```diff
@@ -1,848 +1,972 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
+// 	protoc-gen-go v1.26.0
+// 	protoc        v4.24.3
```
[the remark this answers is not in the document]
FrankYang0529: They are automatically generated.

pkg/imrpc/common.proto
innobead: a note:
> The import path is used to determine which import statements must be generated when one .proto file imports another .proto file. For example, if a.proto imports b.proto, then the generated a.pb.go file needs to import the Go package which contains the generated b.pb.go file (unless both files are in the same package). The import path is also used to construct output filenames. See the "Compiler Invocation" section above for details.

ref: https://protobuf.dev/reference/go/go-generated/

FrankYang0529: The `go_package` has two functions. One is the path for the generated file. Another is package name. The path is not important here is because we generate files in a temporary folder and then move files to current folder.

Based on the temporary folder path in [generate_grpc.sh](https://github.com/longhorn/longhorn-instance-manager/blob/18514f81b9a33b2e146e4e47cc108125627dae66/generate_grpc.sh#L24) script, we copy proto files to `.protobuild/github.com/longhorn/longhorn-manager/pkg/imrpc`. If we use `go_package` like `github.com/longhorn/longhorn-instance-manager/pkg/imrpc`, it will generate files to `.protobuild/github.com/longhorn/longhorn-manager/pkg/imrpc/github.com/longhorn/longhorn-instance-manager/pkg/imrpc`. The path for generated files is different, but the content is similar.

innobead: Discussion with @FrankYang0529 

I think we need to correct the script. It seems to be used the wrong way, even though it works currently. We should just rely on go_package to make things simple.
This is beyond the issue's scope, but it will be good to consolidate the thing here.

innobead (approved):

PR #304, opened 2023-10-27
build: remove unecessary build steps

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/5911


shuo-wu (commented):
These were for some integration tests.

Notice that [this issue](https://github.com/longhorn/longhorn/issues/6743) is revisiting the integration implementation. If we decide to move the engine tests into the current longhorn-instance-manager repo, we cannot remove these 2 dependencies.

ChanYiLin:
Hi @shuo-wu
where can I find the discussion of moving the integration test from longhorn-engine to longhorn-instance-manager repo?

ChanYiLin:
Hi @shuo-wu I have discussed with @FrankYang0529
He tends to move the test to third repo and use submodule to install longhorn-engine and longhorn-instance-manager
So for this PR, we can merge it for now since we don't need them in the future as well.

cc @innobead

shuo-wu (approved):

innobead (approved):

PR #305, opened 2023-11-02
feat(backuptarget): add cleanup all mounts grpc function

ChanYiLin:
ref: [longhorn/longhorn#5741](https://github.com/longhorn/longhorn/issues/5741)

add a function in proxy to unmount the mount points in instance-manager


pkg/imrpc/proxy.proto
```diff
@@ -32,6 +32,8 @@ service ProxyEngineService {
 	rpc SnapshotBackupStatus(EngineSnapshotBackupStatusRequest) returns (EngineSnapshotBackupStatusProxyResponse);
 	rpc BackupRestore(EngineBackupRestoreRequest) returns (EngineBackupRestoreProxyResponse);
 	rpc BackupRestoreStatus(ProxyEngineRequest) returns (EngineBackupRestoreStatusProxyResponse);
+	rpc CleanupAllMounts(google.protobuf.Empty) returns (google.protobuf.Empty);
```
innobead, line 35: Suggest renaming to `CleanupBackupMountPoints`.

If we want to support cleaning up some mount points in the future, you could create a new request message with empty properties instead of using an empty message. This can make the method extensible, even though the current implementation will be just cleaning up all mount points. WDYT?

innobead (changes requested):
Just one feedback.

pkg/imrpc/proxy.proto
ChanYiLin: Sounds good to me, thanks

innobead (approved):

ChanYiLin:
@mergify backport v1.4.x v.1.5x

ChanYiLin:
@mergify backport v1.5.x

PR #306, opened 2023-11-03
Add a linear dm device on the top of a nvme device

derekbit:
longhorn/longhorn#6001

deps on
https://github.com/longhorn/go-spdk-helper/pull/50
https://github.com/longhorn/longhorn-spdk-engine/pull/66


derekbit:
Replaced with https://github.com/longhorn/longhorn-instance-manager/pull/341

PR #307, opened 2023-11-09
Add pprof server in instance-manager pod

derekbit:
Many workloads such as v2 volume backup are executed inside the instance-manager process. To profile the cpu and memory consumptions of the workloads, we can create a pprof server in the instance-manager pod.

Longhorn/longhorn#6377

### DIFF
--- a/app/cmd/start.go
+++ b/app/cmd/start.go
@@ -3,7 +3,10 @@ package cmd
 import (
 	"context"
 	"crypto/tls"
+	"fmt"
 	"net"
+	"net/http"
+	_ "net/http/pprof" // for runtime profiling
 	"os"
 	"os/signal"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


app/cmd/start.go
```diff
@@ -129,6 +132,15 @@ func start(c *cli.Context) (err error) {
 
 	shutdownCh := make(chan error)
 
+	go func() {
+		debugAddress := ":6060"
+		debugHandler := http.DefaultServeMux
+		logrus.Infof("Debug Server listening on %s", debugAddress)
```
shuo-wu, line 138: NIT: `Debug pprof server listening on %s`

shuo-wu (dismissed):
LGTM

app/cmd/start.go
derekbit: Updated. Thank you.

innobead (approved):

PR #315, opened 2023-11-20
build: add code generation script

m-ildefons:
Add a code generation script and build steps in the Dockerfile.dapper This allows the GRPC code generation to happen inside the dapper build environment, which makes it independent of the developers host environment to a greated degree. It is inspired by the equivalent mechanism of the longhorn-engine and helps to avoid headaches when updating GRPC code due to protobuf version discrepancies and similar problems.

related to: https://github.com/longhorn/longhorn/issues/6973


PhanLe1010 (commented):
Are we doing this for engine/instance-manager repo also?

innobead:
Let's revisit this later together with the repos that use gRPC & protos.

PR #316, opened 2023-11-21
feat(mount): request remount when volume mount becomes read only

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/6386

~~- Provide a grpc call to check the volume globalMountPath is readonly or not~~
~~- We can extend the function by simply providing more error message for Controller to check~~

- Instance-Manager will periodically monitor the volume mount point and update the engine status
- Engine/Replica will sync the errorMsg from Instance-Manager
- Volume then check the Engine CRD to get the errorMsg

Pros:
using errorMsg to get the information from process allow us to extend the feature for more situations.


pkg/proxy/volume.go
```diff
@@ -247,3 +253,24 @@ func (p *Proxy) spdkVolumeUnmapMarkSnapChainRemovedSet(ctx context.Context, req
...
 }
+
+func (p *Proxy) VolumeMountStatusCheck(ctx context.Context, req *rpc.VolumeMountStatusRequest) (resp *rpc.VolumeMountStatusResponse, err error) {
+	if req.VolumeName == "" {
+		return nil, grpcstatus.Error(grpccodes.InvalidArgument, "volume name is required")
+	}
+
+	errMsg := ""
+
+	globalMountPath := fmt.Sprintf(types.VolumeGlobalMountPath, sha256.Sum256([]byte(req.VolumeName)))
```
derekbit, line 264: What does it append with a checksum?

ChanYiLin: The global mount path format is as follow
```
/var/lib/kubelet/plugins/kubernetes.io/csi/<csi driver name>/<sha256sum of pv.spec.csi.volumeHandle>/globalmount
```

pkg/process/process_manager.go
```diff
@@ -111,6 +127,60 @@ func (pm *Manager) Shutdown() {
...
+}
+
+func isEngineProcess(p *Process) bool {
+	return p.PortCount == DefaultEnginePortCount
+}
+
+func getVolumeMountPointMap() (map[string]mount.MountPoint, error) {
+	volumeMountPointMap := make(map[string]mount.MountPoint)
+	mounter := mount.New("")
+	mountPoints, _ := mounter.List()
```
c3y1huang, line 169: Suggest not to swallow the error.

ChanYiLin: Fixed, thanks!

ChanYiLin:
Hi @c3y1huang , I have made some change

I move the periodical condition check to another goroutine
I found if I put the periodical check together with broadcast , they might interfere each other and cause some weird behaviors.

Please take a look, thanks!

pkg/process/process_manager.go
```diff
@@ -111,6 +143,64 @@ func (pm *Manager) Shutdown() {
 	close(pm.shutdownCh)
 }
 
+func (pm *Manager) checkMountPointStatusForEngine() {
+	var processToUpdate []*Process
+	volumeMountPointMap, err := getVolumeMountPointMap()
+	if err != nil {
+		logrus.Errorf("Failed to get all volume mount points")
```
c3y1huang, line 150: Should include err.

pkg/process/process_manager.go
```diff
@@ -111,6 +143,64 @@ func (pm *Manager) Shutdown() {
...
+		if isEngineProcess(p) && p.State == StateRunning {
+
+			nameSlices := strings.Split(p.Name, "-")
+			volumeNameSHA := sha256.Sum256([]byte(strings.Join(nameSlices[:len(nameSlices)-2], "-")))
+
+			if mp, exists := volumeMountPointMap[hex.EncodeToString(volumeNameSHA[:])]; exists {
+				p.Conditions[types.EngineConditionFilesystemReadOnly] = false
+				if util.IsMountPointReadOnly(mp) {
+					p.Conditions[types.EngineConditionFilesystemReadOnly] = true
+				}
```
c3y1huang, line 165:
```suggestion
				p.Conditions[types.EngineConditionFilesystemReadOnly] = util.IsMountPointReadOnly(mp)
```

pkg/types/types.go
```diff
@@ -28,3 +28,10 @@ const (
 	InstanceTypeEngine  = "engine"
 	InstanceTypeReplica = "replica"
 )
+
+const (
+	GlobalMountPathPattern = "/host/var/lib/kubelet/plugins/kubernetes.io/csi/driver.longhorn.io/*/globalmount"
+	VolumeGlobalMountPath  = "/host/var/lib/kubelet/plugins/kubernetes.io/csi/driver.longhorn.io/%x/globalmount"
```
c3y1huang, line 34: Could you point to where `VolumeGlobalMountPath` is used?

c3y1huang (changes requested):
> I move the periodical condition check to another goroutine
I found if I put the periodical check together with broadcast , they might interfere each other and cause some weird behaviors.

Any chance for us to get some insights into the weird behaviors?

pkg/types/types.go
ChanYiLin: It is not used in the new implementation, removed, thanks!

ChanYiLin:
Conflict resolved.

pkg/process/process_manager.go
```diff
@@ -104,6 +136,61 @@ func (pm *Manager) startMonitoring() {
...
+	volumeMountPointMap, err := getVolumeMountPointMap()
+	if err != nil {
+		logrus.WithError(err).Warn("Failed to get all volume mount points")
+	}
+
+	pm.lock.RLock()
+	for _, p := range pm.processes {
+		p.lock.Lock()
+		if isEngineProcess(p) && p.State == StateRunning {
+
```
innobead, line 150: Remove unnecessary empty line.

pkg/process/process_manager.go
```diff
@@ -104,6 +136,61 @@ func (pm *Manager) startMonitoring() {
...
+			nameSlices := strings.Split(p.Name, "-")
+			volumeNameSHA := sha256.Sum256([]byte(strings.Join(nameSlices[:len(nameSlices)-2], "-")))
+
+			if mp, exists := volumeMountPointMap[hex.EncodeToString(volumeNameSHA[:])]; exists {
+				p.Conditions[types.EngineConditionFilesystemReadOnly] = util.IsMountPointReadOnly(mp)
+				processToUpdate = append(processToUpdate, p)
+			}
+		}
+		p.lock.Unlock()
```
innobead, line 159: Move just next to `p.lock.Lock()`.

pkg/process/process_manager.go
```diff
@@ -104,6 +136,61 @@ func (pm *Manager) startMonitoring() {
...
+	}
+	pm.lock.RUnlock()
```
innobead, line 161: Move just next to `pm.lock.Lock()`

pkg/process/process_manager.go
```diff
@@ -104,6 +136,61 @@ func (pm *Manager) startMonitoring() {
...
+
+	for _, p := range processToUpdate {
+		p.UpdateCh <- p
+	}
+
+}
+
+func isEngineProcess(p *Process) bool {
+	return p.PortCount == DefaultEnginePortCount
```
innobead, line 170: This is a weak check. It works now, but I suggest differentiating process type. Let's have a ticket for it later.

pkg/util/util.go
```diff
@@ -135,3 +136,12 @@ func IsSPDKTgtReady(timeout time.Duration) bool {
...
 }
+
+func IsMountPointReadOnly(mp mount.MountPoint) bool {
+	for _, opt := range mp.Opts {
+		if opt == "ro" {
+			return true
+		}
+	}
+	return false
+}
```
innobead, line 147: We have a very similar stuff in share manager. I would suggest moving the related code to common-libs, so we can reuse that in all places later. Create a ticket for this later.

[/pkg/server/share_manager.go#L240-L254](https://github.com/longhorn/longhorn-share-manager/blob/de8ff1978f5218656a3a76c3cb6d8ea060d52863/pkg/server/share_manager.go#L240-L254)
```go
func (m *ShareManager) hasHealthyVolume() error {
	mountPath := types.GetMountPath(m.volume.Name)
	if err := exec.CommandContext(m.context, "ls", mountPath).Run(); err != nil {
		return fmt.Errorf(UnhealthyErr, mountPath)
	}

	mounter := mount.New("")
	mountPoints, _ := mounter.List()
	for _, mp := range mountPoints {
		if mp.Path == mountPath && isMountPointReadOnly(mp) {
			return fmt.Errorf(ReadOnlyErr, mountPath)
		}
	}
	return nil
}
```

pkg/process/process_manager.go
```diff
@@ -104,6 +136,61 @@ func (pm *Manager) startMonitoring() {
 	}
 }
 
+func (pm *Manager) checkMountPointStatusForEngine() {
```
innobead, line 139: Suggest adding unit tests, even though it's a private method. This way we can better understand how mount points will be parsed here.

innobead (changes requested):
In general, LGTM. Just a few comments need to resolve.

pkg/process/process_manager.go
ChanYiLin: We also use it in longhorn-maanger: https://github.com/longhorn/longhorn-manager/blob/master/engineapi/instance_manager.go#L290
but I agree we should have more reliable check

pkg/process/process_manager.go
ChanYiLin: we iterate every process
lock it and do some check and unlock it
if we put this next to lock
```
p.lock.Lock()
defer p.lock.Unlock()
```

Then every process will be locked in the end and need to wait until function return to unlock all the process
I think that will effect performance and might stuck some operation?

pkg/util/util.go
ChanYiLin: https://github.com/longhorn/longhorn/issues/7394

pkg/process/process_manager.go
ChanYiLin: https://github.com/longhorn/longhorn/issues/7393

pkg/process/process_manager.go
ChanYiLin: I have move the mount point parsing and process name parsing out of the function and make it simpler to understand. With that, I think we don't need unit tests for it now.
Please take a look, thanks!

pkg/process/process_manager.go
```diff
@@ -104,6 +115,64 @@ func (pm *Manager) startMonitoring() {
 	}
 }
 
+func (pm *Manager) startInstanceConditionCheck() {
```
innobead, line 118: nit: we can merge startInstanceConditionCheck to startMonitor to reduce one goroutine, as it seems not quite necessary to have two here?

pkg/process/process_manager.go
innobead: yeah, misunderstood the scope of defer stack. No need to change here.

innobead (approved):
LGTM.

pkg/process/process_manager.go
innobead: https://github.com/longhorn/longhorn-instance-manager/pull/316#issuecomment-1842403056

PR #333, opened 2023-12-05
Gracefully shut down spdk_tgt

derekbit:
Longhorn/longhorn#7263

deps on https://github.com/longhorn/longhorn-spdk-engine/pull/68


app/cmd/start.go
```diff
@@ -146,89 +154,176 @@ func start(c *cli.Context) (err error) {
 		return err
 	}
 
+	var wg sync.WaitGroup
+
+	waitCount := 4
+	if spdkEnabled {
+		waitCount++
+	}
+	wg.Add(waitCount)
```
innobead, line 163: Suggest removing this, but execute `wg.Add(1)` when needed later.

app/cmd/start.go
```diff
@@ -146,89 +154,176 @@ func start(c *cli.Context) (err error) {
...
+
 	// Start disk server
-	diskGRPCServer, diskGRPCListener, err := setupDiskGRPCServer(diskServiceAddress, spdkServiceAddress, spdkEnabled, shutdownCh)
+	diskGRPCServer, diskGRPCListener, err := setupDiskGRPCServer(diskServiceAddress, spdkServiceAddress, spdkEnabled, shutdown.ch)
 	if err != nil {
 		return err
 	}
 	go func() {
+		defer func() {
```
innobead, line 171: Add `wg.Add(1)` just before the `go func() {`. The concept here is to request when needed, so we don't need to pre-count the number we need for the wait group.

app/cmd/start.go
```diff
@@ -146,89 +154,176 @@ func start(c *cli.Context) (err error) {
...
 		pmGRPCServer.Stop()
+		proxyGRPCServer.Stop()
+		diskGRPCServer.Stop()
 	}()
 
-	return <-shutdownCh
+	wg.Wait()
+
+	if spdkEnabled {
+		stopSPDKTgt(30 * time.Second)
```
innobead, line 291: We start `spdk_tgt` in longhorn-spdk-engine, so can we do stop there as well to make responsibility segregation? The lifecycle of it should be together with the `spdkGRPCServer`. Is this feasible?

pkg/util/util.go
```diff
@@ -135,3 +136,43 @@ func IsSPDKTgtReady(timeout time.Duration) bool {
 	}
 	return false
 }
+
+// FindProcessByCmdline finds the process with matching cmdline
+func FindProcessByCmdline(cmdline string) (*os.Process, error) {
```
innobead, line 141: Does this worth moving to common-lib?

innobead (changes requested):
The concept LG, but having some comments needed to resolve first.

pkg/util/util.go
```diff
@@ -135,3 +136,43 @@ func IsSPDKTgtReady(timeout time.Duration) bool {
...
+	defer procDir.Close()
+
+	processes, err := procDir.Readdir(-1)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, process := range processes {
+		if process.IsDir() {
+			pid, convErr := strconv.Atoi(process.Name())
```
c3y1huang, line 155: Could we utilize [GetProcessPID](https://github.com/longhorn/go-common-libs/blob/main/proc/proc.go#L58)?

app/cmd/start.go
```diff
@@ -146,89 +154,176 @@ func start(c *cli.Context) (err error) {
...
+	}
+
+	logrus.Infof("Stopping spdk_tgt process")
+	if err := process.Signal(syscall.SIGTERM); err != nil {
+		logrus.WithError(err).Errorf("Failed to send SIGTERM to spdk_tgt process")
+	} else {
+		done := make(chan error, 1)
+		go func() {
+			_, err := process.Wait()
+			done <- err
```
c3y1huang, line 313: Should we close the done channel at some point?

app/cmd/start.go
```diff
@@ -146,89 +154,176 @@ func start(c *cli.Context) (err error) {
...
+		}()
+
+		select {
+		case <-time.After(timeout):
+			logrus.Error("spdk_tgt process did not exit in time")
```
c3y1huang, line 318: I am thinking if we should have an additional SIGKILL to forcibly kill the process if it doesn't terminate within the timeout. WDYT?

derekbit: Using `SIGKILL` sounds good.

app/cmd/start.go
derekbit: We start spdk_tgt before starting longhorn-instance-manager ([code](https://github.com/longhorn/longhorn-instance-manager/blob/master/package/instance-manager#L63)). longhorn-spdk-engine ensures the health of spdk_tgt and restarts it if it is somehow crashed.

Therefore, stopping spdk_tgt before exiting longhorn-instance-manager looks reasonable. WDYT?

pkg/util/util.go
derekbit: Yes.

app/cmd/start.go
derekbit: Good catch. Thank you.

app/cmd/start.go
derekbit: Remove WaitGroup. Use errgroup instead.

pkg/util/util.go
derekbit: Addressed. Thank you.

app/cmd/start.go
derekbit: Added. Thank you.

app/cmd/start.go
derekbit: Remove WaitGroup. Use errgroup instead.

app/cmd/start.go
innobead: Sounds good.

c3y1huang (approved):
lgtm

derekbit:
Updated the deps. cc @innobead

innobead (approved):
LGTM

innobead:
@derekbit it's good to have refactoring by replacing the shutdown channel with context to make implementations consistent.

PR #337, opened 2023-12-13
Reconcile engine/replica instance state of v2 volume like v1 volume

derekbit:
https://github.com/longhorn/longhorn/issues/7326


derekbit:
Updated deps and ready to review. cc @innobead

PR #344, opened 2023-12-22
nvme: create a pair of hostid and hostnqn

derekbit:
longhorn/longhorn#7190

deps on https://github.com/longhorn/go-spdk-helper/pull/60

### DIFF
--- a/package/instance-manager
+++ b/package/instance-manager
@@ -25,6 +25,16 @@ function enable_tgtd() {
     tgtd -f 2>&1 | tee /var/log/tgtd.log &
 }
 
+function generate_nvme_hostid_and_hostnqn() {
+    mkdir -p /etc/nvme
+
+    local hostnqn=$(nvme gen-hostnqn)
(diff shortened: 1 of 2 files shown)


innobead (approved):
LGTM

package/instance-manager
```diff
@@ -25,6 +25,16 @@ function enable_tgtd() {
     tgtd -f 2>&1 | tee /var/log/tgtd.log &
 }
 
+function generate_nvme_hostid_and_hostnqn() {
+    mkdir -p /etc/nvme
+
+    local hostnqn=$(nvme gen-hostnqn)
+    # hostnqn is generated from /sys/class/dmi/id/product_uuid according to the implemetation of libnvme
+    echo "$hostnqn" > /etc/nvme/hostnqn
```
innobead, line 33: I assume `/etc/nvme` exists or just make -p to ensure it?

derekbit: `/etc/nvme` doesn't exist because we compile nvme-cli in the `builder` container and copy it into the `release` container.

innobead (approved):

PR #358, opened 2024-01-05
proto: regenerate files using correct versions

derekbit:

Issue #

proto: regenerate files using correct versions


FrankYang0529 (approved):
LGTM. Thank you.

PR #364, opened 2024-01-10
Make the best effort to clean up stale NVMe and device mapper devices

derekbit:
No need to error out.

Issue Longhorn/longhorn#7608

Just make the best effort to clean up stale NVMe and device mapper devices.

### DIFF
--- a/app/cmd/start.go
+++ b/app/cmd/start.go
@@ -140,19 +140,18 @@ func cleanupStaledNvmeAndDmDevices() error {
 	for _, sys := range subsystems {
 		logrus.Infof("Found NVMe subsystem %+v", sys)
 		if strings.HasPrefix(sys.NQN, helpertypes.NQNPrefix) {
-			logrus.Infof("Cleaning up NVMe subsystem %v: NQN %v", sys.Name, sys.NQN)
-
-			if err := helpernvme.DisconnectTarget(sys.NQN, executor); err != nil {
-				return errors.Wrapf(err, "failed to disconnect NVMe subsystem %v: NQN %v", sys.Name, sys.NQN)
-			}
-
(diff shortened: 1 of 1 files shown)


app/cmd/start.go
```diff
@@ -143,7 +143,7 @@ func cleanupStaledNvmeAndDmDevices() error {
 			logrus.Infof("Cleaning up NVMe subsystem %v: NQN %v", sys.Name, sys.NQN)
 
 			if err := helpernvme.DisconnectTarget(sys.NQN, executor); err != nil {
-				return errors.Wrapf(err, "failed to disconnect NVMe subsystem %v: NQN %v", sys.Name, sys.NQN)
+				logrus.WithError(err).Warnf("Failed to disconnect NVMe subsystem %v: NQN %v", sys.Name, sys.NQN)
```
shuo-wu, line 146: NIT: "...will continue the cleanup"

app/cmd/start.go
```diff
@@ -152,7 +152,7 @@ func cleanupStaledNvmeAndDmDevices() error {
 			}
 			logrus.Infof("Removing dm device %v", dmDeviceName)
 			if err := helperutil.DmsetupRemove(dmDeviceName, true, true, executor); err != nil {
-				return errors.Wrapf(err, "failed to remove dm device %v", dmDeviceName)
+				logrus.WithError(err).Warnf("Failed to remove dm device %v", dmDeviceName)
```
shuo-wu, line 155: NIT: "...will continue the cleanup"

shuo-wu (commented):
Should we clean up the dm device first?

derekbit:
> Should we clean up the dm device first?

Yes. Update the order, but it is not harmful in this case. Thank you.

shuo-wu (approved):

PR #369, opened 2024-01-11
Refactor instance gRPC methods

derekbit:

Issue longhorn/longhorn#7628

Refactor instance gRPC methods and make them easy to maintain.


pkg/instance/instance.go
```diff
@@ -17,35 +17,58 @@ import (
...
 
 const (
 	maxMonitorRetryCount     = 10
 	monitorRetryPollInterval = 1 * time.Second
 )
 
-type Server struct {
-	ctx                          context.Context
-	logsDir                      string
+type InstanceOps interface {
```
innobead, line 31: nit: support `InstanceWatch` later.

innobead (approved):

PR #370, opened 2024-01-11
Refactor disk gRPC methods

derekbit:

Issue https://github.com/longhorn/longhorn/issues/7628


derekbit:
Will update longhorn-manager after this ticket gets merged.

innobead (approved):

PR #377, opened 2024-01-22
feat(mount): process manager periodically check mount status and upda…

ChanYiLin:

Since there are grpc update and many refactors not backport
So for this feature, we have to manually backport it.


ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7501

PR #378, opened 2024-01-22
feat(mount): process manager periodically check mount status and update engine status

ChanYiLin:
Since there are grpc update and some refactors not backport
So for this feature, we have to manually backport it.


ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7501

innobead (changes requested):

innobead:
@ChanYiLin will create another new PR from his fork instead.

PR #382, opened 2024-01-23
Support TLS in proxy clients

sfackler:
Issue longhorn/longhorn#7040

The proxy client was never updated to support mTLS, breaking all communication through it when the longhorn-grpc-tls secret was present.

The proxy gRPC API doesn't provide any endpoints that respond directly without proxying through to another process. Since we will need one to check if we can use TLS or not in longhorn-manager I just used the gRPC health API instead. Happy to add another endpoint to the proxy API if you'd prefer though.

See https://github.com/sfackler/longhorn-manager/commit/248b0c622495b35f2b22e765fd78dd4fabb9ffeb for the longhorn-manager change.

I've tested this fix backported to 1.5.3 locally and it resolves the issues in longhorn/longhorn#7040


pkg/client/proxy.go
```diff
@@ -61,15 +65,19 @@ type ProxyClient struct {
...
 			grpc.WithKeepaliveParams(keepalive.ClientParameters{
 				Time:                time.Second * 10,
 				PermitWithoutStream: true,
 			}),
 		}
+		if tlsConfig != nil {
+			dialOptions = append(dialOptions, grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)))
+		} else {
+			dialOptions = append(dialOptions, grpc.WithTransportCredentials(insecure.NewCredentials()))
+		}
```
ejweber, line 80: All other clients use the `util.Connect` helper method (now) to handle this. I prefer to do so here as well, unless there is a strong reason not to. It looks like the main differences are:

- We need to handle some strangeness around the missing protocol (`tcp://`) that `util.Connect` expects.
- `util.Connect` uses a 30 second keepalive instead of the 10 second keepalive established by @c3y1huang in https://github.com/longhorn/longhorn-instance-manager/pull/119. I suspect the 30 second timeout is fine here, but if not, we could make it configurable in `util.Connect`.

pkg/client/proxy.go
```diff
@@ -154,3 +172,9 @@ func (c *ProxyClient) ClientVersionGet() (version emeta.VersionOutput) {
 	logrus.Trace("Getting client version")
 	return emeta.GetVersion()
 }
+
+func (c *ProxyClient) CheckConnection() error {
+	req := &healthpb.HealthCheckRequest{}
+	_, err := c.health.Check(getContextWithGRPCTimeout(c.ctx), req)
+	return err
+}
```
ejweber, line 180: This seems like a good solution to me. We already have the health check endpoint available in the server, so we may as well make use of it (instead of adding a new API).

ejweber (changes requested):
LGTM in general. One change I'd like to see if possible.

I also reviewed the proposed changes at https://github.com/sfackler/longhorn-manager/commit/248b0c622495b35f2b22e765fd78dd4fabb9ffeb#diff-98e9e49f06725ee91e72520d23fde8521949c662f852777a797358a604d5340c. They look correct to me. They are mostly directly copied from https://github.com/longhorn/longhorn-manager/blob/7f189c0affedac7ef98c41a01d4620b068d18791/engineapi/instance_manager.go#L142-L158 (which I think is a good thing). We will need a PR with those changes as well, but that PR must import this one. It looks safe to merge this one asynchronously when ready.

pkg/client/proxy.go
sfackler: Updated! It looks like `uti.Connect` should work with a raw `host:port` address since `parseEndoint` has a fallback when there's no `://` in the input.

ejweber:
This is looking good to me from the code side.

Testing https://github.com/longhorn/longhorn/issues/3839#issuecomment-1121849339 with:

- `webericjames/longhorn-instance-manager:a0c0fa299379f9a9783ff5512c90a7defdac516c` - I build an image of this PR and pushed it directly.
- `webericjames/longhorn-manager:556c40ae3-dirty` - I modified https://github.com/sfackler/longhorn-manager/commit/248b0c622495b35f2b22e765fd78dd4fabb9ffeb to incorporate his latest instance-manager changes and rebased it to pick up some necessary changes from master.

Unfortunately, I need a bit more time, as I ran into some certificate issues originally.

pkg/client/proxy.go
```diff
@@ -40,6 +39,7 @@ type ServiceContext struct {
 	quit context.CancelFunc
 
 	service rpc.ProxyEngineServiceClient
+	health  healthpb.HealthClient
```
derekbit, line 42: Off the topic:
Just notice the clients of process manager and instance don't have `heath`.
Is it better to add it into the two clients as well?
cc @ejweber @innobead

derekbit (approved):
The implementation LGTM.

@sfackler Will you file a PR in longhorn-manager side later?

pkg/client/proxy.go
derekbit: @c3y1huang 
Can you elaborate more on the increase of the 30-seconds keepalive ping time? The default value is 10 seconds.

c3y1huang: I don't recall why I used 10 seconds, might just be something that looks fitting at the time. Using 30 seconds keepalive looks fine to me.

derekbit:
> This is looking good to me from the code side.
> 
> Testing [longhorn/longhorn#3839 (comment)](https://github.com/longhorn/longhorn/issues/3839#issuecomment-1121849339) with:
> 
> * `webericjames/longhorn-instance-manager:a0c0fa299379f9a9783ff5512c90a7defdac516c` - I build an image of this PR and pushed it directly.
> * `webericjames/longhorn-manager:556c40ae3-dirty` - I modified [sfackler/longhorn-manager@248b0c6](https://github.com/sfackler/longhorn-manager/commit/248b0c622495b35f2b22e765fd78dd4fabb9ffeb) to incorporate his latest instance-manager changes and rebased it to pick up some necessary changes from master.
> 
> Unfortunately, I need a bit more time, as I ran into some certificate issues originally. My main concern is a flood of scary looking logs while running the test cases. I need to confirm whether I am doing something wrong or there are further changes we need to make.
> 
> Seeing a lot of these:
> 
> ```
> [longhorn-manager-z785d] W0124 23:13:46.915518       1 logging.go:59] [core] [Channel #3 SubChannel #4] grpc: addrConn.createTransport failed to connect to {Addr: "10.42.2.2:8500", ServerName: "longhorn-backend.longhorn-system", }. Err: connection error: desc = "transport: authentication handshake failed: tls: first record does not look like a TLS handshake"
> [longhorn-manager-jrjz6] W0124 23:13:47.646924       1 logging.go:59] [core] [Channel #3 SubChannel #4] grpc: addrConn.createTransport failed to connect to {Addr: "10.42.5.32:8500", ServerName: "longhorn-backend.longhorn-system", }. Err: connection error: desc = "transport: authentication handshake failed: tls: first record does not look like a TLS handshake"
> [longhorn-manager-hmc47] W0124 23:13:47.654869       1 logging.go:59] [core] [Channel #3 SubChannel #4] grpc: addrConn.createTransport failed to connect to {Addr: "10.42.1.13:8500", ServerName: "longhorn-backend.longhorn-system", }. Err: connection error: desc = "transport: authentication handshake failed: tls: first record does not look like a TLS handshake"
> ```
> 
> 8500 is the `InstanceManagerProcessManagerServiceDefaultPort`, so I think this may not be related to the current change (which deals with the proxy service), but need to investigate more. (The system works while this is happening, so it seems we must be falling back to non-tls communication.)
> 
> @sfackler, do you see these longhorn-manager logs in your testing?

@ejweber 
I remove the unused connection to process manager gRPC server in https://github.com/longhorn/longhorn/issues/7783.
You can please take a look. Thank you.

c3y1huang (approved):
LGTM

pkg/client/proxy.go
derekbit: > I don't recall why I used 10 seconds, might just be something that looks fitting at the time. Using 30 seconds keepalive looks fine to me.

You use 30 seconds, but the default value is 10 seconds. Just curious about why you increased it to 30 seconds.

pkg/client/proxy.go
innobead: If there is a healthy server implemented. this sounds good to me.

innobead (approved):
LGTM. 

Wait for @ejweber verification and review first before merging this PR.

sfackler:
> @sfackler, do you see these longhorn-manager logs in your testing?

I didn't check logs when I was testing it out unfortunately - I just verified that volumes attached and the UI worked.

ejweber (approved):
LGTM. I still want to do some sanity testing with @derekbit's PR (https://github.com/longhorn/longhorn-manager/pull/2527) to confirm this is working well enough as a whole to ship, but I don't anticipate needing any more changes here.

pkg/client/proxy.go
ejweber: To clarify (unless I'm missing something), @joshimoo made the decision to go with 30 seconds in https://github.com/longhorn/longhorn-instance-manager/pull/109 for other services. @c3y1huang chose 10 seconds (the minimum) in https://github.com/longhorn/longhorn-instance-manager/pull/119 for the proxy service.

derekbit: > To clarify (unless I'm missing something), @joshimoo made the decision to go with 30 seconds in #109 for other services. @c3y1huang chose 10 seconds (the minimum) in #119 for the proxy service.

Ah, yes. Sorry for the misunderstanding. BTW, any pros and cons for the bigger/smaller value?

ejweber: I am not too sure. If I understand correctly, a smaller value can help us detect a broken connection sooner during a long RPC (especially a streaming one). Since the server might not send back any data for a long time, the keepalive helps us know whether it is still trying.

On the other hand, the gRPC docs mention to avoid configuring keepalive much below one minute for DDoS reasons. That seems unlikely to be a major consideration for us since there are a limited number of clients.

https://github.com/longhorn/longhorn-instance-manager/blob/a0c0fa299379f9a9783ff5512c90a7defdac516c/vendor/google.golang.org/grpc/clientconn.go#L194-L210

innobead:
@mergify backport v1.6.x v1.5.x v1.4.x

PR #392, opened 2024-02-07
feat(volume): add remount read only volume grpc

ChanYiLin:
ref: longhorn/longhorn 7843


pkg/types/types.go
```diff
@@ -45,6 +45,7 @@ const (
 
 const (
 	GlobalMountPathPattern = "/host/var/lib/kubelet/plugins/kubernetes.io/csi/driver.longhorn.io/*/globalmount"
+	GlobalMountPathFormat  = "/host/var/lib/kubelet/plugins/kubernetes.io/csi/driver.longhorn.io/%s/globalmount"
```
shuo-wu, line 48: It just occurred to me that Longhorn supports customized kubelet root dir. Then is this global mount path still valid?

pkg/proxy/volume.go
```diff
@@ -308,3 +314,18 @@ func (ops V2DataEngineProxyOps) VolumeSnapshotMaxSizeSet(ctx context.Context, re
...
 }
+
+func (p *Proxy) RemountReadOnlyVolume(ctx context.Context, req *rpc.RemountVolumeRequest) (resp *emptypb.Empty, err error) {
+	volumeName := req.VolumeName
+	volumeNameSHA := sha256.Sum256([]byte(volumeName))
+	volumeNameSHAStr := hex.EncodeToString(volumeNameSHA[:])
+
+	volumeGlobalMountPath := fmt.Sprintf(types.GlobalMountPathFormat, volumeNameSHAStr)
+
+	cmd := exec.CommandContext(ctx, "mount", "-o", "remount,rw", volumeGlobalMountPath)
```
shuo-wu, line 325: Do we need to use `NamespaceExecutor` in `"github.com/longhorn/go-common-libs/ns"` instead?

pkg/types/types.go
ChanYiLin: Yeah you are right, let me see if we can fix it.

pkg/proxy/volume.go
ChanYiLin: It is manipulating the host directory, I think there is no need to use `NamespaceExecutor`
Is it correct?

pkg/types/types.go
ChanYiLin: Changed to use regex to build `volumeNameSHA->mountPoint` map
and then use volumeNameSHA to get the mountPoint and get the `mp.Path`
Then we don't need to know the custom kubelet directory

pkg/proxy/volume.go
shuo-wu: This command is executed in the host namespace rather than the current container namespace. I think it should use `NamespaceExecutor`. This is similar to `iscsiadm` cmds.

ChanYiLin: Fixed, thanks!

shuo-wu (dismissed):

proto/vendor/protobuf/src/google/protobuf/empty.proto
```diff
@@ -40,7 +40,7 @@ option objc_class_prefix = "GPB";
 option csharp_namespace = "Google.Protobuf.WellKnownTypes";
 option cc_enable_arenas = true;
 
-// A generic empty message that you can re-use to avoid defining duplicated
+// A generic empty message that you can reuse to avoid defining duplicated
```
innobead, line 43: This file is not what we controlled, so no need to fix the codespell issue here. I updated codespell config to skip proto folder.

innobead (changes requested):
LGTM, just one request.

proto/vendor/protobuf/src/google/protobuf/empty.proto
ChanYiLin: Fixed, thanks!

innobead (approved):

ChanYiLin:
@mergify backport v1.6.1

ChanYiLin:
@mergify backport v1.6.x

PR #393, opened 2024-02-07
[BACKPORT v1.5.x] feat(volume): remount volume if it becomes read only

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7843


ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7844

ChanYiLin:
We didn't introduce go common lib in v1.5.x, v1.4.x,
I will remain the same implementation for commaned execution
cc @innobead

pkg/proxy/volume.go
```diff
@@ -247,3 +253,24 @@ func (p *Proxy) spdkVolumeUnmapMarkSnapChainRemovedSet(ctx context.Context, req
...
+	volumeNameSHA := sha256.Sum256([]byte(volumeName))
+	volumeNameSHAStr := hex.EncodeToString(volumeNameSHA[:])
+
+	volumeMountPointMap, err := util.GetVolumeMountPointMap()
+	if err != nil {
+		logrus.WithError(err).Warn("Failed to get all volume mount points")
+	}
+
+	if mp, exists := volumeMountPointMap[volumeNameSHAStr]; exists {
+		cmd := exec.CommandContext(ctx, "mount", "-o", "remount,rw", mp.Path)
```
shuo-wu, line 268: We should use the ns executor for v1.5.x as well.

innobead: @ChanYiLin  Let's make sure we have done the testing as well since the change is different from 1.6 and master after using go-common-libs.

innobead: This is a global mount. Should we still need to use nsenter here?

ChanYiLin: Besides, here in our CSI part
https://github.com/longhorn/longhorn-manager/blob/master/csi/util.go#L295
```
isMnt, err := mounter.IsMountPoint(path)
	if os.IsNotExist(err) {
		return false, os.MkdirAll(path, 0750)
	}

	IsCorruptedMnt := mount.IsCorruptedMnt(err)
```
we create the mount path (both globalpath and pod path) directly from os.MkdirAll
and use mount to check the mount path is valid

innobead: cc @derekbit @c3y1huang @shuo-wu having any concerns?

shuo-wu: Discussed with @ChanYiLin. It seems that `mountPropagation: Bidirectional` option allows the mount point inside the CSI plugin to be propagated to the host, and the similar logic applies to IM pod and the host when the IM pod uses **bidirectional** mount for host. The only exception is [the NFS mount case](https://github.com/longhorn/longhorn-manager/blob/a568043026841264d9c98d9ca5a228fb94774eca/csi/node_server.go#L815-L818).

As for the NFS case, the remount handled by the IM pod will overwrite the CSI plugin `nsenter` mount. I am not sure if the new remount info (in the IM pod) still work fine for the below cases:
1. The CSI plugin pod crash.
2. The share manager pod crash. Is the backend-recover service still able to read the new mount info?
3. The instance manager pod crash. 

@derekbit

derekbit: Backend-recover service only records the connection state including client hostname and ip, so I think it is nothing to do with the mount/remount option.

ChanYiLin: I have tested it with rwx volume, and remount still works

shuo-wu: Then I am fine with the current implementation

shuo-wu (approved):

innobead (approved):

PR #394, opened 2024-02-07
[BACKPORT v1.4.x] feat(volume): remount volume if it becomes read only

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7843


ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7845

ChanYiLin:
We didn't introduce go common lib in v1.5.x, v1.4.x,
I will remain the same implementation for commaned execution
cc @innobead

innobead (approved):

PR #402, opened 2024-02-18
Support setting spdk_tgt log level and flags

derekbit:

Issue https://github.com/longhorn/longhorn/issues/7939

Deps on https://github.com/longhorn/longhorn-spdk-engine/pull/108


innobead (approved):

pkg/proxy/volume.go
```diff
@@ -336,7 +336,7 @@ func (p *Proxy) RemountReadOnlyVolume(ctx context.Context, req *rpc.RemountVolum
 			"remount,rw",
 			mp.Path,
 		}
-		if _, err := nsexec.Execute("mount", opts, lhtypes.ExecuteDefaultTimeout); err != nil {
+		if _, err := nsexec.Execute(nil, "mount", opts, lhtypes.ExecuteDefaultTimeout); err != nil {
```
innobead, line 339: @ChanYiLin @derekbit already fixed the issue here.

pkg/meta/version.go
```diff
@@ -2,7 +2,7 @@ package meta
 
 const (
 	// InstanceManagerAPIVersion is used for compatibility check for longhorn-manager
-	InstanceManagerAPIVersion    = 5
+	InstanceManagerAPIVersion    = 6
```
innobead, line 5: Although we add new methods to `InstanceService` gRPC interface, it does not really need to bump `InstanceManagerAPIVersion` here, because:

- When an old client calls any existing methods in a new instance manager w/ these new methods, everything should work correctly because of backward compatibility in gRPC.
- When a new client calls any new method in an old instance manager, even though the method does not exist, the call will just do nothing, because of forward compatibility.

We Need to review the current traditional version flag way to control how new/old longhorn manager communicates with new/old instance manager, but rather rely on gRPC capabilities.

ref: https://github.com/longhorn/longhorn/issues/4833

PR #418, opened 2024-03-05
refactor(utils): move is mount read only function to common lib

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7394

Add mount related util function: IsMountPointReadOnly
This is used in share-manager and instance-manager to detect if the mount point becomes readonly.

Note:
1. Since go-common-libs had some breaking (function parameters not the same) changes in this PR
So when updating the version, we also need to fix some code to meet the changes in go-common-libs

e.g.
```
if _, err := nsexec.Execute(nil, "mount", opts, lhtypes.ExecuteDefaultTimeout); err != nil {}
```

2. since go-common-libs has been updated, 
    - `go-spdk-helper` is also needed to update 
    - `longhorn-spdk-engine` is also needed to update

cc @derekbit

PR #428, opened 2024-03-07
[MANUALLY BACKPORT v1.5.x] refactor(utils): move is mount read only function to common lib

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7394

Add mount related util function: IsMountPointReadOnly
This is used in share-manager and instance-manager to detect if the mount point becomes readonly.


innobead (approved):

ChanYiLin:
ref: https://github.com/longhorn/longhorn-instance-manager/pull/418

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8125

PR #433, opened 2024-03-11
Lh7960 update dependencies and resolve breaking changes

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7960

Since longhorn-engine updates `go-common-libs`/`go-iscsi-helper`
We need to update longhorn-engine in longhonr-instance-manager's vendor

Dependency PR:
- `longhorn-engine`: https://github.com/longhorn/longhorn-engine/pull/1050


ChanYiLin:
Wrong target branch

PR #434, opened 2024-03-11
chore(vendor): update go-common-libs/go-iscsi-utils and resolve breaking changes

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7960

Since longhorn-engine updates `go-common-libs`/`go-iscsi-helper`
We need to update longhorn-engine in longhonr-instance-manager's vendor

Dependency PR:
- `longhorn-engine`: https://github.com/longhorn/longhorn-engine/pull/1050


go.mod
```diff
@@ -66,3 +65,5 @@ require (
 )
 
 replace golang.org/x/text v0.3.2 => golang.org/x/text v0.3.3
+
+replace github.com/longhorn/longhorn-engine v1.5.4 => github.com/ChanYiLin/longhorn-engine v1.4.0-rc1.0.20240311091244-3365ed4916a7
```
innobead, line 69: This is not updated yet?

innobead (commented):
If this PR is not ready, should mark as draft.

go.mod
ChanYiLin: Sure I can change it to Draft
I use to mention what this PR is depended on in the comment above

> Dependency PR:
longhorn-engine: https://github.com/longhorn/longhorn-engine/pull/1050

innobead:
@ChanYiLin Let's merge this after the confit is fixed. Thanks.

innobead (approved):

PR #435, opened 2024-03-12
refactor: remove engine binary path check

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8094

The original check limit the pattern of the engine binary path.
But there is no need to check the engine binary path.
Even the path is wrong, we catch the error and show the error message.

### DIFF
--- a/pkg/process/process_manager.go
+++ b/pkg/process/process_manager.go
@@ -4,7 +4,6 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
-	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


ChanYiLin:
coretest: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6621/console

innobead (approved):
LGTM.

innobead:
Out of PR, I noticed that when a process fails to create, it is not unregistered. Could you please create a ticket to address this issue, @ChanYiLin? Thanks.

[/pkg/process/process_manager.go#L254-L268](https://github.com/longhorn/longhorn-instance-manager/blob/dce3a124861eb6b2a3c9f08115dfe40657e52326/pkg/process/process_manager.go#L254-L268)
```go
	if err := pm.registerProcess(p); err != nil {
		return nil, err
	}

	p.UpdateCh <- p
	if err := p.Start(); err != nil {
		// initializing failed so we sent event about the failed state, but still return the process rpc below
		// this is to be consistent with the prior implementation
		logrus.WithError(err).Errorf("Process Manager: failed to init new process %v", req.Spec.Name)
		p.UpdateCh <- p
	} else {
		logrus.Infof("Process Manager: created process %v", req.Spec.Name)
	}

	return p.RPCResponse(), nil
```

derekbit (approved):
LGTM

innobead:
@mergify backport v1.6.x

PR #437, opened 2024-03-12
fix(process): unregister process when it fails to start in creation

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8155

### DIFF
--- a/pkg/process/process_manager.go
+++ b/pkg/process/process_manager.go
@@ -217,6 +217,7 @@ func (pm *Manager) ProcessCreate(ctx context.Context, req *rpc.ProcessCreateRequ
 		// this is to be consistent with the prior implementation
 		logrus.WithError(err).Errorf("Process Manager: failed to init new process %v", req.Spec.Name)
 		p.UpdateCh <- p
+		pm.unregisterProcess(p)
 	} else {
 		logrus.Infof("Process Manager: created process %v", req.Spec.Name)
 	}


ChanYiLin:
**TEST RUNNING**

https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6623/console

ChanYiLin:
**TEST PASSED**
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6623/console

shuo-wu:
I am thinking if this is necessary. In our current implementation including process/v2volume/backingimage creation, we will blindly cache the item no matter if the creation succeeds or not. Once the creation fails, the cached record will tell the detailed info & error message when callers try to check it. And the callers need to clean up the error records in an appropriate manner.

ChanYiLin:
Yes @shuo-wu you are right 
1. When the process fails to start, it marks the process as error 
https://github.com/longhorn/longhorn-instance-manager/blob/6372baca91084d36461ea860f8e7b45115a1a858/pkg/process/process.go#L53-L58

2. It update the process status to Error
https://github.com/longhorn/longhorn-instance-manager/blob/6372baca91084d36461ea860f8e7b45115a1a858/pkg/process/process_manager.go#L257-L261

3. in longhorn volume controller it will then update it to desired state Stop
https://github.com/longhorn/longhorn-manager/blob/edf23eddc3b6c307031eaa770c3d312c963f25a5/controller/volume_controller.go#L728-L739

4. The process will then be deleted in instance reconcile
https://github.com/longhorn/longhorn-manager/blob/edf23eddc3b6c307031eaa770c3d312c963f25a5/controller/instance_handler.go#L359-L385

 
So we shouldn't unregister it from process manager, or the process get won't be able to get it.

innobead:
100% right, well done @shuo-wu and thanks @ChanYiLin to clarify the context.

PR #444, opened 2024-03-19
Dockerfile: Update SPDK commit ID

shuo-wu:

Issue https://github.com/longhorn/longhorn/issues/8187

We forgot to update the SPDK commit ID for instance manager after doing that in longhorn/longhorn-spdk-engine


shuo-wu:
@mergify backport v1.6.x

innobead (approved):

innobead:
cc @longhorn/dev-data-plane

PR #450, opened 2024-03-22
Dockerfile: Update SPDK commit ID

DamiaSan:
With this commit id the new branch longhorn-v24.01 will be used.


shuo-wu (approved):
Please update `Dockerfile.dapper` for `longhorn-spdk-engine` and `go-spdk-helper` as well

derekbit (approved):

PR #451, opened 2024-03-22
Allow retrying killing a process when it get stucks in stopping state

shuo-wu:

https://github.com/longhorn/longhorn/issues/8091

We don't know how to reproduce this issue. But this PR allows retrying the process kill after a timeout anyway

### DIFF
--- a/pkg/process/process.go
+++ b/pkg/process/process.go
@@ -161,8 +161,11 @@ func (p *Process) StopWithSignal(signal syscall.Signal) {
 			}
 		}()
 
-		if cmd == nil || !cmd.Started() {
-			logrus.Errorf("Process Manager: cmd of %v hasn't started, no need to stop", p.Name)
+		if cmd == nil || !cmd.IsRunning() {
+			logrus.Errorf("Process Manager: cmd of %v is not running anymore, no need to stop", p.Name)
+			if p.State != StateStopped && p.State != StateError {
+				p.State = StateStopped
+			}
 			return
 		}
 
--- a/pkg/process/process.go
+++ b/pkg/process/process.go
@@ -29,12 +29,13 @@ type Process struct {
 	PortCount int32
 	PortArgs  []string
 
-	UUID       string
-	State      State
-	ErrorMsg   string
-	Conditions map[string]bool
-	PortStart  int32
-	PortEnd    int32
+	UUID              string
+	State             State
+	ErrorMsg          string
+	Conditions        map[string]bool
+	PortStart         int32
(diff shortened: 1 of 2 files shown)


pkg/process/process.go
```diff
@@ -161,22 +174,29 @@ func (p *Process) StopWithSignal(signal syscall.Signal) {
...
-				return
+		if needStop {
+			logrus.Infof("Process Manager: trying to stop process %v", p.Name)
+			cmd.StopWithSignal(signal)
+			for i := 0; i < types.WaitCount; i++ {
+				if p.IsStopped() {
+					return
+				}
+				logrus.Infof("Wait for process %v to shutdown", p.Name)
+				time.Sleep(types.WaitInterval)
```
ejweber, line 194: If I understand correctly, we think:

- The goroutine that SHOULD have successfully stopped the process got stuck here and could never finish?
- Another goroutine like this one could never be created because `p.State` was `StateStopping`.

And the crux of this intermediate solution is to allow another goroutine to be created after `DeleteTimeout` that skips straight to the kill step?

If my understanding is correct, my only suggestion is to not hard code `DeleteTimeout` directly and instead set it to `WaitInterval * WaitCount`. This way the new behavior only takes over after it seems the old expected behavior has failed.

ejweber (commented):
IIUC, we think the user from the original issue

pkg/process/process.go
```diff
@@ -137,15 +138,26 @@ func (p *Process) Stop() {
...
+	needStop, needTimeoutKill := false, false
+	now := time.Now()
+
 	p.lock.Lock()
 	if p.State != StateStopping && p.State != StateStopped && p.State != StateError {
 		p.State = StateStopping
+		p.DeletionTimestamp = &now
 		needStop = true
 	}
+	// Retry the deletion if the process is not stopped in 30 seconds
```
derekbit, line 150: 60 seconds?

WaitInterval is 100ms and WaitCount is 600.

innobead (approved):

derekbit (approved):
LGTM

innobead:
@mergify backport v1.6.x 1.5.x

innobead:
cc @longhorn/qa need to ensure replica related e2e test cases succeed as usual.

innobead:
@mergify backport v1.5.x

pkg/process/process.go
shuo-wu: Yeah. I forgot to update this comment. Will do it with a new PR

PR #461, opened 2024-04-09
Add gRPC validation

shuo-wu:

We can remove this when https://github.com/FrankYang0529/longhorn-interface is introduced officially. The code in this PR is similar to that of longhorn-engine.


innobead:
cc @FrankYang0529

PR #474, opened 2024-04-20
feat(upgrade): support v2 volume upgrade

derekbit:

Issue longhorn/longhorn#6001

### DIFF
--- a/pkg/client/instance.go
+++ b/pkg/client/instance.go
@@ -88,6 +88,9 @@ func NewInstanceServiceClientWithTLS(ctx context.Context, ctxCancel context.Canc
 type EngineCreateRequest struct {
 	ReplicaAddressMap map[string]string
 	Frontend          string
+	InitiatorAddress  string
+	TargetAddress     string
+	UpgradeRequired   bool
 }
 
 type ReplicaCreateRequest struct {
@@ -115,6 +118,7 @@ type InstanceCreateRequest struct {
 	BackendStoreDriver string
 }
 
+// InstanceCreate creates an instance.
 func (c *InstanceServiceClient) InstanceCreate(req *InstanceCreateRequest) (*api.Instance, error) {
 	if req.Name == "" || req.InstanceType == "" {
 		return nil, fmt.Errorf("failed to create instance: missing required parameter")
@@ -169,6 +173,10 @@ func (c *InstanceServiceClient) InstanceCreate(req *InstanceCreateRequest) (*api
 
 			ProcessInstanceSpec: processInstanceSpec,
 			SpdkInstanceSpec:    spdkInstanceSpec,
+
+			UpgradeRequired:  req.Engine.UpgradeRequired,
+			InitiatorAddress: req.Engine.InitiatorAddress,
+			TargetAddress:    req.Engine.TargetAddress,
 		},
 	})
 	if err != nil {
@@ -178,6 +186,7 @@ func (c *InstanceServiceClient) InstanceCreate(req *InstanceCreateRequest) (*api
 	return api.RPCToInstance(p), nil
 }
 
+// InstanceDelete deletes the instance by name.
 func (c *InstanceServiceClient) InstanceDelete(dataEngine, name, instanceType, diskUUID string, cleanupRequired bool) (*api.Instance, error) {
 	if name == "" {
 		return nil, fmt.Errorf("failed to delete instance: missing required parameter name")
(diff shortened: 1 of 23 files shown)


pkg/instance/instance.go
```diff
@@ -702,11 +712,170 @@ func engineResponseToInstanceResponse(e *spdkapi.Engine) *rpc.InstanceResponse {
...
+
+	ops, ok := s.ops[req.DataEngine]
+	if !ok {
+		return nil, grpcstatus.Errorf(grpccodes.Unimplemented, "unsupported data engine %v", req.DataEngine)
+	}
+	return ops.InstanceDeleteTarget(req)
+}
+
+func (ops V1DataEngineInstanceOps) InstanceDeleteTarget(req *rpc.InstanceDeleteTargetRequest) (*emptypb.Empty, error) {
+	return nil, grpcstatus.Error(grpccodes.Unimplemented, "v1 data engine instance target switch over is not supported")
```
DamiaSan, line 859: "v1 data engine instance target delete is not supported"

DamiaSan (changes requested):

pkg/instance/instance.go
derekbit: Updated, thank you.

derekbit:
@DamiaSan @shuo-wu  Can you help review the PR? Thank you.

cc @innobead

pkg/client/instance.go
```diff
@@ -115,6 +118,17 @@ type InstanceCreateRequest struct {
 	BackendStoreDriver string
 }
 
+type InstanceSuspendRequest struct {
```
c3y1huang, line 121: Could you point to where this is used?

DamiaSan (dismissed):

pkg/client/instance.go
derekbit: It is dead code. Removed. Thank you.

DamiaSan (approved):

pkg/instance/instance.go
```diff
@@ -702,11 +722,170 @@ func engineResponseToInstanceResponse(e *spdkapi.Engine) *rpc.InstanceResponse {
...
+
+	switch req.Type {
+	case types.InstanceTypeEngine:
+		err := c.EngineSwitchOverTarget(req.Name, req.TargetAddress)
+		if err != nil {
+			return nil, grpcstatus.Error(grpccodes.Internal, errors.Wrapf(err, "failed to switch over target for engine %v", req.Name).Error())
+		}
+		return &emptypb.Empty{}, nil
+	case types.InstanceTypeReplica:
+		return nil, grpcstatus.Errorf(grpccodes.InvalidArgument, "target switchover is not supported for instance type %v", req.Type)
```
c3y1huang, line 848:
```suggestion
		return nil, grpcstatus.Errorf(grpccodes.InvalidArgument, "target switch over is not supported for instance type %v", req.Type)
```

pkg/instance/instance.go
```diff
@@ -702,11 +722,170 @@ func engineResponseToInstanceResponse(e *spdkapi.Engine) *rpc.InstanceResponse {
...
+		return nil, grpcstatus.Errorf(grpccodes.InvalidArgument, "unknown instance type %v", req.Type)
+	}
+}
+
+func (s *Server) InstanceDeleteTarget(ctx context.Context, req *rpc.InstanceDeleteTargetRequest) (*emptypb.Empty, error) {
+	logrus.WithFields(logrus.Fields{
+		"name":       req.Name,
+		"type":       req.Type,
+		"dataEngine": req.DataEngine,
+	}).Info("Delete target")
```
c3y1huang, line 859:
```suggestion
	}).Info("Deleting target")
```

c3y1huang (changes requested):

pkg/instance/instance.go
derekbit: Updated. Thank you.

pkg/instance/instance.go
derekbit: Updated. Thank you.

c3y1huang (approved):
Do you think it would be beneficial to separate the (ops V1DataEngineInstanceOps) and (ops V2DataEngineInstanceOps) into different files? Maybe something to consider for future refactoring.

derekbit:
> Do you think it would be beneficial to separate the (ops V1DataEngineInstanceOps) and (ops V2DataEngineInstanceOps) into different files? Maybe something to consider for future refactoring.

Yes, same thought. I will improve it after v1.7.0.

PR #477, opened 2024-04-24
Add fsFreeze field to VolumeSnapshot

ejweber:

longhorn/longhorn#2187

- Proxy the freezeFS field for SnapshotVolume through to the engine.
- Attempt to unfreeze all file systems on the root partitions of Longhorn volumes on startup.

Depends on:

- https://github.com/longhorn/types/pull/8
- https://github.com/longhorn/longhorn-engine/pull/1083

Need to fix imports before merging.

### DIFF
--- a/app/cmd/start.go
+++ b/app/cmd/start.go
@@ -24,11 +24,13 @@ import (
 	healthpb "google.golang.org/grpc/health/grpc_health_v1"
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/reflection"
+	"k8s.io/mount-utils"
 
 	commonTypes "github.com/longhorn/go-common-libs/types"
 	helpernvme "github.com/longhorn/go-spdk-helper/pkg/nvme"
 	helpertypes "github.com/longhorn/go-spdk-helper/pkg/types"
 	helperutil "github.com/longhorn/go-spdk-helper/pkg/util"
+	engineutil "github.com/longhorn/longhorn-engine/pkg/util"
 	spdk "github.com/longhorn/longhorn-spdk-engine/pkg/spdk"
 	spdkutil "github.com/longhorn/longhorn-spdk-engine/pkg/util"
 	rpc "github.com/longhorn/types/pkg/generated/imrpc"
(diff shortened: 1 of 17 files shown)


PhanLe1010:
> - Proxy the freezeFS field for SnapshotVolume through to the engine.
> - Attempt to unfreeze all file systems on the root partitions of Longhorn volumes on startup.

Is it correct that we are not going to periodically check and unfreeze the filesystem of a crashed engine like:
* instance-manager pod is still running 
* engine crashed 
* engine move to new node
* instance-manager of old node is not going to unfreeze the filesystem on old node

ejweber:
> Is it correct that we are not going to periodically check and unfreeze the filesystem of a crashed engine like:

It is correct (unless there is pushback of course). This is my opinion on why:

- The engine doesn't actually usually "crash". Usually it transitions to an error state because (for example) it lost all connections with a replica. Then, it is shut down. We do implement shutdown logic that will ensure it is unfrozen in this case.
- If the engine literally does go away with any shut down logic (e.g. with a `kill -9`), it is usually automatically reattached by Longhorn. We do implement logic that will ensure it is unfrozen on startup.

In my opinion, the complexity of periodically scanning for and attempting to unfreeze file systems for engines that happen to somehow crash without going through any shutdown at the exact moment they are frozen introduces more risk than the potential for this situation to leave the file system frozen does. (Especially in the case where we have multiple instance managers running simultaneously, as still often happens for v1.)

Keep in mind that if it does happen, it is not nearly as serious as the other cases we investigated. With a look at the support bundle, we can determine the file system was frozen, but never unfrozen, and suggest an unfreeze command.

PhanLe1010:
Agree! Multiple instance-manager case introduces significant complications if we introduce the periodically check and unfreeze in IM

james-munson (dismissed):
LGTM.

james-munson (dismissed):
No issues.

PhanLe1010 (dismissed):
LGTM

app/cmd/start.go
```diff
@@ -183,6 +208,10 @@ func start(c *cli.Context) (err error) {
 		if err := cleanupStaledNvmeAndDmDevices(); err != nil {
 			return err
 		}
+	} else {
+		if err := unfreezeFilesystems(); err != nil {
+			return err
+		}
```
shuo-wu, line 219: Can we do this during IM startup? As we discussed before, after the upgrade, will the new IM pod wrongly unfreeze fs handled by old IM pods?

ejweber: It will. However, I think the benefits outweigh the risks:

- Since snapshots are very fast, there will likely not be any ongoing in the old instance manager during the restart.
- If there ARE freezing file systems in another instance manager, we cannot stop their freeze, but we might unfreeze them immediately after it completes. This means their data will be flushed, but the snapshot MIGHT be taken after some I/O manages to resume. This is exactly the same behavior we have today, so the situation only improves with the feature addition.
- The worst thing we can do with this feature (IMO) is leave froze file systems stranded on nodes. It will lead to unhappy users and extra support engagements.

shuo-wu: OK. Got it! I agree that the benefits outweigh the risks, and things won't be worse than the today's implementation

shuo-wu (dismissed):

ejweber:
Hello @PhanLe1010, @shuo-wu, and @james-munson. I have resolved the conflict. Can I have an approve again from one of you so I can merge? Thanks!

james-munson (approved):

PR #479, opened 2024-04-27
Fix that instance-manager is stuck at starting state

derekbit:

Issue Longhorn/longhorn#8455

Add BaseDelay to grpc.ConnectParams.Backoff.


innobead (approved):

innobead:
Is this due to the recent change only impacting the master?

derekbit:
> Is this due to the recent change only impacting the master?

Yes, it is a regression introduced from [fix: golangci-lint error](https://github.com/longhorn/longhorn-instance-manager/commit/5ae373f1c4d781bce22287a158d9c80b5343f0f5).
It is not backported to other branches yet.

innobead:
@mergify backport v1.6.x

PR #486, opened 2024-04-30
ci: change to use GitHub actions

FrankYang0529:
https://github.com/longhorn/longhorn/issues/8039

Sample build: https://github.com/FrankYang0529/longhorn-instance-manager/actions/runs/8892845609


.github/workflows/build.yml
```diff
@@ -0,0 +1,213 @@
...
+
+    - name: Upload binaries
+      uses: actions/upload-artifact@v4
+      with:
+        name: binaries_amd64_artifact
+        path: ./bin/*
+
+  build-arm64-binaries:
+    name: Build ARM64 binaries
+    runs-on: ["self-hosted", "ARM64"]
```
innobead, line 36: Use label `oracle-aarch64-2cpu-8gb` instead.

.github/workflows/build.yml
```diff
@@ -0,0 +1,213 @@
...
+        push: true
+        platforms: linux/amd64
+        tags: longhornio/longhorn-instance-manager:${{ github.ref_name }}-amd64
+        file: package/Dockerfile
+        build-args: |
+          ARCH=amd64
+
+  build-push-arm64-images:
+    name: Build and push ARM64 images
+    runs-on: ["self-hosted", "ARM64"]
```
innobead, line 115: Fix the label.

innobead (changes requested):
LGTM, but need to fix the arm64 runner label first.

.github/workflows/build.yml
FrankYang0529: I think we have to deploy ARM runner ourselves. The `oracle-aarch64-2cpu-8gb` can only be used in branch head. If we set it, the pull requests can't really run CI. For example, I used `oracle-aarch64-2cpu-8gb` in https://github.com/longhorn/go-spdk-helper/pull/105, but the job doesn't run.

FrankYang0529: Update to use CNCF runners. Thanks.

innobead (approved):

FrankYang0529:
@mergify backport v1.6.x

FrankYang0529:
@mergify backport v1.5.x

PR #491, opened 2024-05-07
Fix high CPU usage of instance-manager pod caused by for loop

derekbit:

Issue Longhorn/longhorn#8496


innobead:
Good fix. Unfortunately, we missed that in the previous review.

PhanLe1010:
@mergify backport 1.6.x

PR #498, opened 2024-05-13
vendor: update dependencies

derekbit:

Issue Longhorn/longhorn#8551


derekbit:
> LGTM, and only needs to track if there are any side effects with updating `aws-sdk-go` to v1.52.2

Thanks @mantissahz 
Let's check if there is any side effect by regression test.

PR #510, opened 2024-06-07
Dockerfile: verify the dependencies for the binaries

derekbit:

Issue Longhorn/longhorn#8063

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -131,13 +131,16 @@ COPY --from=builder \
     /usr/sbin/
 
 COPY --from=builder \
-    /usr/local/lib64 \
-    /usr/local/lib64
+   /usr/local/lib64 \
+   /usr/local/lib64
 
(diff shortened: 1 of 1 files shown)


innobead (approved):

package/Dockerfile
```diff
@@ -131,14 +131,17 @@ COPY --from=builder \
...
+   /usr/local/lib64 \
+   /usr/local/lib64
 
 RUN ldconfig
 
 COPY package/bin/longhorn-instance-manager /usr/local/bin/
 COPY package/instance-manager /usr/local/bin/
 
+# Verify the dependencies for the binaries
+RUN ldd /usr/local/bin/* /usr/local/sbin/* /usr/sbin/* | grep "not found" && exit 1 || true
```
innobead, line 143: It's better to have `grep -i` for case-insensitive filtering.

PR #512, opened 2024-06-11
Make long gRPC timeout configurable

PhanLe1010:
longhorn/longhorn#2765

Note: we need to update go.mod to have this changes: before merging:
1. https://github.com/longhorn/types/pull/14
2. https://github.com/longhorn/longhorn-engine/pull/1125


innobead (approved):
LGTM. Merge this until go.mod is updated.

PhanLe1010:
I think we should merge in the order longhorn-engine -> instance-manager -> longhorn-manager since the latter is depended on the former

pkg/client/proxy.go
```diff
@@ -108,8 +109,14 @@ func getContextWithGRPCTimeout(parent context.Context) context.Context {
...
 }
 
-func getContextWithGRPCLongTimeout(parent context.Context) context.Context {
-	ctx, _ := context.WithTimeout(parent, GRPCServiceLongTimeout)
+// getContextWithGRPCLongTimeout returns a context with given grpcTimeoutSeconds + GRPCServiceTimeout timeout.
+// If grpcTimeoutSeconds is 0, use the default GRPCServiceLongTimeout instead.
+func getContextWithGRPCLongTimeout(parent context.Context, grpcTimeoutSeconds int64) context.Context {
+	grpcTimeout := GRPCServiceLongTimeout
+	if grpcTimeoutSeconds > 0 {
+		grpcTimeout = (time.Second * time.Duration(grpcTimeoutSeconds)) + GRPCServiceTimeout
```
derekbit, line 117: Can we use `grpcTimeout = max(grpcTimeoutSeconds, GRPCServiceTimeout)` instead?

ejweber: It looks like this is here to mimic:

https://github.com/longhorn/longhorn-instance-manager/blob/0c7a73712efea62b9a465983f2ec0d353d6c0bd4/pkg/client/proxy.go#L101-L104

The context for this is https://github.com/longhorn/longhorn/issues/4096. I am not sure I understand the details, but it seems it is intentional for the timeout here to be longer than the desired timeout at the engine layer.

ejweber (commented):
LGTM in general, though I'm interested in any context you have on the timeout addition @derekbit mentioned.

pkg/client/proxy.go
PhanLe1010: @ejweber is correct, we want to have slight bigger timeout in the proxy client side comparing to the actual timeout in the engine/replica because proxy server add some delay to the flow

innobead:
This PR is actually together with the async revision counter mechanism, which is better to have an individual commit to make it obvious.

innobead (dismissed):

go.mod
```diff
@@ -6,22 +6,24 @@ require (
...
 	golang.org/x/sync v0.7.0
-	google.golang.org/grpc v1.63.2
-	google.golang.org/protobuf v1.34.1
+	google.golang.org/grpc v1.64.0
+	google.golang.org/protobuf v1.34.2
 	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c
 	k8s.io/mount-utils v0.30.1
 )
 
+replace github.com/longhorn/longhorn-engine => ../longhorn-engine
```
innobead, line 25: After updating, we can merge this PR.

pkg/client/proxy.go
derekbit: I see. @PhanLe1010 Can you leave a comment here for better understanding the context? Thanks.

PhanLe1010: Surely. Fixed. Thank you

go.mod
PhanLe1010: PR updated. Thank you

derekbit (approved):
LGTM

PR #528, opened 2024-06-22
Dockerfile: add procps package for missing ps utility

derekbit:
The instance-manager pod for V2 volume is being terminated due to a failed liveness probe in https://github.com/longhorn/longhorn-manager/blob/master/controller/instance_manager_controller.go#L1314-L1315.
```
		processProbe := "[ $(ps aux | grep 'spdk_tgt' | grep -v 'grep' | grep -v 'tee' | wc -l) != 0 ]"
		livenessProbes = append(livenessProbes, processProbe)
```

This failure is caused by the removal of the ps utility (procps package) in the recent BCI 15.5 and 15.6 images.

Longhorn 8807

Issue #


innobead (approved):

derekbit:
@Mergify backport v1.6.x

PR #534, opened 2024-06-27
feat(snapshot): use snapshot timestamp as created

DamiaSan:

Issue # 7641

We must use the new attribute SnapshotTimestamp, instead CreationTime, as Created info into v2 SnapshotList.

I have also removed the snapshot options fields into `EngineSnapshotCreate` because they are not needed, user created and snapshot timestamp fields are all managed inside `longhorn-spdk-engine`.


pkg/process/process_manager.go
```diff
@@ -19,7 +19,8 @@ import (
 	"google.golang.org/protobuf/types/known/emptypb"
 	"k8s.io/mount-utils"
 
-	lhUtils "github.com/longhorn/go-common-libs/utils"
+	lhKubernetes "github.com/longhorn/go-common-libs/kubernetes"
+	lhLonghorn "github.com/longhorn/go-common-libs/longhorn"
 	"github.com/longhorn/longhorn-instance-manager/pkg/types"
 	"github.com/longhorn/longhorn-instance-manager/pkg/util"
 	"github.com/longhorn/longhorn-instance-manager/pkg/util/broadcaster"
```
derekbit, line 26:
```suggestion
	lhKubernetes "github.com/longhorn/go-common-libs/kubernetes"
	lhLonghorn "github.com/longhorn/go-common-libs/longhorn"

	"github.com/longhorn/longhorn-instance-manager/pkg/types"
	"github.com/longhorn/longhorn-instance-manager/pkg/util"
	"github.com/longhorn/longhorn-instance-manager/pkg/util/broadcaster"
```

derekbit (approved):

PR #535, opened 2024-06-28
chore(dockerfile): fix the dependencies of spdk_tgt missing caused by bci-base:15.6

derekbit:

Issue Longhorn/longhorn#8846


derekbit:
Merge the ticket to prevent hindering tickets for testing.
cc @innobead @shuo-wu @DamiaSan

PR #555, opened 2024-07-15
fix(dockerfile): fix spdk_tgt hang and exit on arm64 platform after enabling debug mode

derekbit:

Issue Longhorn/longhorn#9004

Disable the debug mode first. The issue doesn't always happen on all arm64 platforms and needs further investigation.


innobead (approved):

innobead:
Need to check the build failure. Why did we need fuse-devel when building SPDK tgt? If we don't use it, is it feasible to skip it while building?

derekbit:
> Why did we need fuse-devel when building SPDK tgt? 

Default spdk_tgt needs fuse-devel. I disabled it before because we don't need it. To avoid any surprise like the debug mode on arm64 platform, I think we can retain it which doesn't have negative impacts.

derekbit:
> Need to check the build failure. 

All passed and ready for review. cc @innobead

derekbit:
@mergify backport v1.7.x

PR #576, opened 2024-07-24
fix(snapshot): set user created to true for v2 engine

DamiaSan:
This fix resolve an issue in 1.6.2, where all snapshots have erroneously the user created flag set to false. So, when upgrading to 1.7.0, all these snapshots can't be restored.

Issue #9066


pkg/proxy/snapshot.go
```diff
@@ -16,6 +16,7 @@ import (
 	spdktypes "github.com/longhorn/longhorn-spdk-engine/pkg/types"
 
 	"github.com/longhorn/longhorn-instance-manager/pkg/util"
+	"github.com/longhorn/longhorn-spdk-engine/pkg/api"
 
 	rpc "github.com/longhorn/longhorn-instance-manager/pkg/imrpc"
```
derekbit, line 21:
```suggestion

	spdktypes "github.com/longhorn/longhorn-spdk-engine/pkg/types"
	"github.com/longhorn/longhorn-spdk-engine/pkg/api"

	"github.com/longhorn/longhorn-instance-manager/pkg/util"

	rpc "github.com/longhorn/longhorn-instance-manager/pkg/imrpc"
```

derekbit (approved):

innobead (approved):

PR #625, opened 2024-08-13
feat(bitmap): use enhanced bitmap from go-common-libs

james-munson:

https://github.com/longhorn/longhorn/issues/7802

Switch to use the enhanced bitmap utility to avoid immediate reuse and port collision.


james-munson:
The big part of this PR is switch from local bitmap to an enhanced one to be added to go-common-libs in https://github.com/longhorn/go-common-libs/pull/60

Marking it as `Draft` until that one is committed and the `replace =>` in go.mod can be removed.

DamiaSan (approved):
Just a little remainder for us all: in this repo the packages from `go-common-libs` are imported as `lh`, meanwhile in `longhorn-spdk-engine` they are imported as `common`. For example we have `lhBitmap` and `commonBitmap` for the same package, maybe in the future we could align them.

derekbit (approved):

james-munson:
@mergify backport v1.7.x

james-munson:
I would like to see this fix backported to 1.7.2, so I have set a PR up for that.

PR #700, opened 2024-10-28
fix(logging): demote some Info logging to Debug

james-munson:

Part of https://github.com/longhorn/longhorn/issues/6702

Demote some noisy Info-level logging to Debug. Routine snapshotting can flood logs with Info-level output.

### DIFF
--- a/pkg/proxy/snapshot.go
+++ b/pkg/proxy/snapshot.go
@@ -24,7 +24,7 @@ func (p *Proxy) VolumeSnapshot(ctx context.Context, req *rpc.EngineVolumeSnapsho
 		"volumeName": req.ProxyEngineRequest.VolumeName,
 		"dataEngine": req.ProxyEngineRequest.DataEngine,
 	})
-	log.Infof("Snapshotting volume: snapshot %v", req.SnapshotVolume.Name)
+	log.Debugf("Snapshotting volume: snapshot %v", req.SnapshotVolume.Name)
 
 	ops, ok := p.ops[req.ProxyEngineRequest.DataEngine]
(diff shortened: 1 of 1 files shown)


PhanLe1010 (approved):
LGTM

Btw, these logs has been very useful when debugging users/customer issue in the past. It would be more difficult if we don't have them. But I guess in a prod, we still shouldn't have too many logs

derekbit:
> LGTM
> 
> Btw, these logs has been very useful when debugging users/customer issue in the past. It would be more difficult if we don't have them. But I guess in a prod, we still shouldn't have too many logs

These log messages only print once for each operation and won't result in flooding the logs, so I suggest keeping them at the INFO level.

derekbit:
> > LGTM
> > Btw, these logs has been very useful when debugging users/customer issue in the past. It would be more difficult if we don't have them. But I guess in a prod, we still shouldn't have too many logs
> 
> These log messages only print once for each operation and won't result in flooding the logs, so I think we should keep them at the INFO level.

WDYT? @PhanLe1010 @innobead @james-munson

innobead:
Agreed with @derekbit @PhanLe1010 . I think we should just keep this at the original level and not make panics later for troubleshooting.

shuo-wu:
Agreed as well. keeping these logs at INFO level definitely benefits troubleshooting.

innobead:
Let's close this PR first as discussed. cc @james-munson

james-munson:
These log statements were picked based on the fact that each of them appeared more than 1000 times over the course of a two-week period in one support bundle.

PR #716, opened 2024-11-19
feat(v2 upgrade): support engine live upgrade

derekbit:
Signed-off-by: Derek Su <derek.su@suse.com>#### Which issue(s) this PR fixes:

Issue Longhorn/longhorn#9104


DamiaSan (approved):

pkg/client/proxy_replica.go
```diff
@@ -204,12 +204,15 @@ func (c *ProxyClient) ReplicaRemove(dataEngine, serviceAddress, engineName, repl
...
 		"serviceAddress": serviceAddress,
 		"engineName":     engineName,
-		"replicaAddress": replicaAddress,
 	}
 	if err := validateProxyMethodParameters(input); err != nil {
 		return errors.Wrap(err, "failed to remove replica for volume")
 	}
 
+	if replicaAddress == "" && replicaName == "" {
+		return fmt.Errorf("failed to remove replica for volume: replica address and name are both empty")
```
DamiaSan, line 213: Maybe can be useful to log the volume name too.

derekbit: longhorn-manager will receive the error and log the error with the volume name, so we don't need to add the volume name here.

DamiaSan (approved):

PR #726, opened 2024-11-26
fix(logging): separate methods for internal and SPDK logging setting

james-munson:

Issue https://github.com/longhorn/longhorn/issues/6702

Separate the setting of instance-manager log level from that for v2 data engine (spdk_tgt).


derekbit (approved):
LGTM

james-munson:
@mergify backport v1.7.x v1.6.x

PR #770, opened 2024-12-17
fix(v2): introduce instance manager v2 prestop script

c3y1huang:

Issue longhorn/longhorn#9959

Introduce instance manager v2 pre-stop script

`None`

`None`


c3y1huang:
@coderabbitai review

derekbit (commented):
Is it possible to skip the `cleanup_spdk_resources` if the ndoe is being live upgrade?
For example, if node.spec.liveUpgrade is true, we don't need to do cleanup_spdk_resources().

c3y1huang:
> Is it possible to skip the `cleanup_spdk_resources` if the ndoe is being live upgrade? For example, if node.spec.liveUpgrade is true, we don't need to do cleanup_spdk_resources().

After discussing with @derekbit , we considered three approaches:
1. **Environment variable to control the pre-stop script:** This approach didn't work in my tests. Ref:
    - https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks
    - https://github.com/kubernetes/kubernetes/issues/90624
1. **File-based control:** For example adding this to the pre-stop script:
    ```bash
    [[ -f "/tmp/skip-instance-manager-v2-prestop" ]] && echo "Skipping SPDK resources cleanup" || cleanup_spdk_resources
    ```
    This approach works as expected.
1. **Kubernetes API to retrieve CR details:** While this could be feasible, it's not ideal since we want to keep the instance manager independent of Kubernetes awareness.

Based on the above, yes we could skip the pre-stop script cleanup, and IMO, the file-based approach might be most ideal.

PR #831, opened 2025-01-14
chore: versions of dependent packages are centrally managed in longhorn/dep-versions

derekbit:

Issue Longhorn/longhorn#10208

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -2,6 +2,7 @@
 FROM  registry.suse.com/bci/golang:1.23 AS gobuilder
 
 ARG ARCH=amd64
+ARG DEP_VERSION_FILE=versions.json
 
 RUN zypper -n ref && \
     zypper update -y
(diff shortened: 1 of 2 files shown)


innobead (commented):
Have we considered building a versioned lib image when updating the dep version file (that should be release version based) and the image includes all necessary dependent libraries (source code or prebuilt), so it can be just used here and it's not required to repeat the process of pulling and building code?

package/Dockerfile
```diff
@@ -55,28 +51,36 @@ RUN zypper -n addrepo --refresh https://download.opensuse.org/repositories/syste
...
 # Build TGT
-RUN cd /usr/src && \
-    git clone https://github.com/rancher/tgt.git && \
+RUN export TGT_REPO=$(jq -r '.["tgt"].repo' /versions.json) && \
+    export TGT_COMMIT_ID=$(jq -r '.["tgt"].commit' /versions.json) && \
+    cd /usr/src && \
+    git clone ${TGT_REPO} && \
     cd tgt && \
     git checkout ${TGT_COMMIT_ID} && \
     make; \
```
c3y1huang, line 75:
```suggestion
    make && \
```

derekbit:
> Have we considered building a versioned lib image when updating the dep version file (that should be release version based) and the image includes all necessary dependent libraries (source code or prebuilt), so it can be just used here and it's not required to repeat the process of pulling and building code?

Yes, we can design the flow. One concern is the versioned lib image might make the development more time-consuming.
For example, RD updates longhorn-spdk-engine and longhorn-instance-manager, then it needs to build a full lib image and a IM image.

derekbit:
> Have we considered building a versioned lib image when updating the dep version file (that should be release version based) and the image includes all necessary dependent libraries (source code or prebuilt), so it can be just used here and it's not required to repeat the process of pulling and building code?

Yes, we can design the flow. One concern is the versioned lib image might make the development more time-comsumping.

package/Dockerfile
derekbit: Fixed. Thank you.

derekbit:
> Have we considered building a versioned lib image when updating the dep version file (that should be release version based) and the image includes all necessary dependent libraries (source code or prebuilt), so it can be just used here and it's not required to repeat the process of pulling and building code?

My original plan is to build an RPM package for each library and install them into a BCI-micro image while building the Longhorn component image as AppCo. However, there are some challenges:

- When to build the RPM packages
- Where to store the RPM packages
- Don't make the development and debugging complicated

innobead:
> Yes, we can design the flow. One concern is the versioned lib image might make the development more time-consuming.
> 
> For example, RD updates longhorn-spdk-engine and longhorn-instance-manager, then it needs to build a full lib image and a IM image.  

I thought each should be independent, because the libs defined in https://github.com/longhorn/dep-versions/blob/main/versions.json are all library based? Which parts are missing?

Btw, I am also fine with this implementation.

derekbit:
> > Yes, we can design the flow. One concern is the versioned lib image might make the development more time-consuming.
> > For example, RD updates longhorn-spdk-engine and longhorn-instance-manager, then it needs to build a full lib image and a IM image.

Got it. What you mentioned is like https://github.com/longhorn/longhorn-instance-manager/pull/831#issuecomment-2633005105.

scripts/package
```diff
@@ -39,7 +39,44 @@ do
...
+  local branch=$(convert_version_to_major_minor_x "$version")
+
+  echo "${branch}"
+}
+
+BRANCH=$(get_branch)
+
+wget -q "https://raw.githubusercontent.com/longhorn/dep-versions/refs/heads/${BRANCH}/versions.json" -O package/versions.json || {
+  echo "Failed to fetch from branch $BRANCH, trying main branch..."
+  wget -q "https://raw.githubusercontent.com/longhorn/dep-versions/refs/heads/main/versions.json" -O package/versions.json || {
```
james-munson, line 68: In longhorn-manager/master branch, `get_branch` will return "v1.9.x", a branch that does not exist in longhorn-manager or in dep-versions repo.  Is that why there is a failover to dep-version/head branch here?  It will also use dep-versions/main to build backport branches of longhorn-manager until the dep-version branches are created with their own versions.json.  (Which will be pretty soon, I would guess.)

james-munson (approved):
Yes, that looks good.  Just one question.

scripts/package
derekbit: yes, you are right!

c3y1huang (approved):

derekbit:
@mergify backport v1.8.x

PR #853, opened 2025-02-06
feat: new makefile target for secure image build

COLDTURNIP:
#### Which issue(s) this PR fixes:

Supports security build.

#### What this PR does / why we need it:

Prepare for potential secure release

#### Special notes for your reviewer:

The secrets provided by EIO MUST NOT present in any visible output.

#### Additional documentation or context

Guidance on achieving SLSA Level 3

PR #884, opened 2025-02-17
fix(ci): fix binary dependency for local image packaging

COLDTURNIP:

Issue #853

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -119,10 +119,6 @@ jobs:
       run: |
         chmod +x ./bin/*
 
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -119,10 +119,6 @@ jobs:
       run: |
         chmod +x ./bin/*
 
-    - name: Copy bin folder to package
-      run: |
-        cp -r ./bin ./package/
-
```
COLDTURNIP, line 125: Collect binaries from ./bin directly.

derekbit (approved):

PR #898, opened 2025-02-25
[BACKPORT][v1.8.1] chore(vendor): update dependencies

derekbit:

Issue #

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/longhorn/backupstore v0.0.0-20250215022153-7ee3a979f2a9
 	github.com/longhorn/go-common-libs v0.0.0-20250215052214-151615b29f8e
 	github.com/longhorn/go-spdk-helper v0.0.0-20250116051812-eabe34a4219e
-	github.com/longhorn/longhorn-engine v1.8.0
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/longhorn/backupstore v0.0.0-20250215022153-7ee3a979f2a9
 	github.com/longhorn/go-common-libs v0.0.0-20250215052214-151615b29f8e
 	github.com/longhorn/go-spdk-helper v0.0.0-20250116051812-eabe34a4219e
-	github.com/longhorn/longhorn-engine v1.8.0
+	github.com/longhorn/longhorn-engine v1.9.0-dev-20241215.0.20250225130735-e63c1afeb26b
```
c3y1huang, line 12: Shouldn't this be v1.8.1-dev-... instead?

derekbit: It should be...the commit id is correct, but the `go mod tidy` adds the version v1.9.0-dev.
https://github.com/longhorn/longhorn-engine/commits/v1.8.x/

c3y1huang (approved):

go.mod
derekbit: Ref: [go pseudo-versions](https://go.dev/ref/mod#pseudo-versions)

PR #899, opened 2025-02-26
feat: implement MetricsGet to support V1 and V2 data engines in ProxyOps

hookak:
https://github.com/longhorn/longhorn/issues/10472


derekbit (approved):
LGTM

PR #910, opened 2025-03-12
chore(Dockerfile): revert registry.suse.com/bci/bci-base to 15.6 and use latest golangci-lint

derekbit:

Issue longhorn/longhorn#10573

- chore(Dockerfile): revert registry.suse.com/bci/bci-base to 15.6
- chore(Dockerfile): use latest golangci-lint
- chore(Dockerfile): retry zypper commands if they are failed

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -8,7 +8,7 @@ ARG SRC_TAG
 RUN zypper -n ref && \
     zypper update -y
 
-RUN zypper -n addrepo --refresh https://download.opensuse.org/repositories/network:utilities/SLE_15_SP5/network:utilities.repo && \
+RUN zypper -n addrepo --refresh https://download.opensuse.org/repositories/network:utilities/SLE_15/network:utilities.repo && \
     zypper --gpg-auto-import-keys ref
 
 RUN zypper -n install wget jq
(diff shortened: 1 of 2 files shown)


package/Dockerfile
```diff
@@ -36,7 +38,7 @@ RUN export GRPC_HEALTH_PROBE_DOWNLOAD_URL=$(wget -qO- https://api.github.com/rep
     chmod +x /usr/local/bin/grpc_health_probe
 
 # Stage 2: build binary from c source code
-FROM registry.suse.com/bci/bci-base:15.7 AS cbuilder
+FROM registry.suse.com/bci/bci-base:15.6 AS cbuilder
```
c3y1huang, line 41: How does reverting `bci-base` to `15.6` help resolving the instability issue?

derekbit: Reverting bci-base to 15.6 is due to it is a Tech Preview version.
For the stability, it is addressed in https://github.com/longhorn/longhorn-instance-manager/pull/910/commits/f266aee1fcb5194f9703de08ef6f46af8cbb30bc.

innobead:
The context about using the latest golangci-lint. 

ref: https://golangci-lint.run/welcome/faq/
> The same as the Go team (the 2 latest minor versions).
> Basically, golangci-lint supports Go versions lower or equal to the Go version used to compile it.
> New versions of Go are not automatically supported because, in addition of the Go version used to build it, some linters and/or internal pieces of golangci-lint could need to be adapted to support this new Go version.

PR #948, opened 2025-04-15
fix: remove unnecessary commands and add gpg key verification back

PhanLe1010:
1. No need to snappy repo for the build container
2. No need to printout the build script
3. Need to verify gpg key when installing package

longhorn/longhorn#9456


derekbit (approved):
LGTM

PR #988, opened 2025-05-14
feat: return UUID in instance Get and List operations

derekbit:

Issue Longhorn/longhorn#10888


pkg/instance/instance.go
```diff
@@ -246,6 +249,10 @@ func (ops V2DataEngineInstanceOps) InstanceDelete(req *rpc.InstanceDeleteRequest
 		}
 	}()
 
+	if req.Uuid != "" {
+		logrus.Debugf("Deleting instance %v with UUID %v. UUID is ignored in V2 data engine.", req.Name, req.Uuid)
```
derekbit, line 253: No need to mention v2 data engine is not supported in the data plane.

```suggestion
		logrus.Debugf("Deleting instance %v with UUID %v", req.Name, req.Uuid)
```

derekbit (commented):
In general, LGTM.

c3y1huang (approved):

derekbit:
@mergify backport v1.9.x

PR #1020, opened 2025-07-03
fix: update ReplicaRebuildStatus field name for type compatibility

hookak:

https://github.com/longhorn/longhorn-manager/pull/3736

This change is required to support the build for the following PR: https://github.com/longhorn/longhorn-manager/pull/3736


derekbit (approved):
LGTM

PR #1049, opened 2025-08-14
fix: allow removal of v2 volume head's parent

yulken:

Issue #9064


shuo-wu:
Besides, for this feature, we need to check if there is any similar check in the longhorn-manager repo as well.

PR #1069, opened 2025-09-05
feat: introduce new options to intance-manager for no hugepage support

derekbit:

Issue Longhorn/longhorn#7066


c3y1huang (approved):

derekbit:
> Should we also update the help menu?

Good catch!

derekbit:
> Should we also update the help menu?

Updated. Thank you.

c3y1huang (approved):

PR #1086, opened 2025-09-15
feat: gracefully shut down spdk_tgt and update disk client

derekbit:

Issue Longhorn/longhorn#11760

### DIFF
--- a/pkg/api/disk.go
+++ b/pkg/api/disk.go
@@ -13,6 +13,7 @@ type DiskInfo struct {
 	FreeBlocks  int64
 	BlockSize   int64
 	ClusterSize int64
+	State       string
 }
 
 // ReplicaStorageInstance is utilized to represent a replica directory of a legacy volume and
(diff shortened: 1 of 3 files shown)


package/exec-logrotate
```diff
@@ -17,6 +17,14 @@ function create_logrotate_config() {
...
     copytruncate
 }
+
+/log/prestop-${DATA_ENGINE}.log {
+    rotate 2
+    size 20M
+    missingok
+    notifempty
+    copytruncate
+}
```
shuo-wu, line 27: Hmm... Do we need logrotate for the cleanup script? I am thinking it's a one-time job.

derekbit: We append the prestop messages to the log file, so it is better to rotate.

shuo-wu: I mean, do we need to do rotation? Anyway, this does not hurt anything. Let me approve it first

shuo-wu (approved):

mantissahz (approved):
LGTM

PR #1385, opened 2026-05-11
fix: implement V2 proxy for VolumeSnapShotMaxCountSet/SizeSet

carterli0407-cell:
The V2 stubs in pkg/proxy/volume.go previously returned Unimplemented.
Wire them to call the actual SPDK client methods.

Issue #12921


derekbit:
In favor of https://github.com/longhorn/longhorn-instance-manager/pull/1391

PR #1410, opened 2026-05-22
fix: wait for spdk_tgt process to terminate during pre-stop cleanup

derekbit:

Issue Longhorn/longhorn#13179

### DIFF
--- a/package/instance-manager-v2-prestop
+++ b/package/instance-manager-v2-prestop
@@ -24,6 +24,26 @@ function cleanup_spdk_resources() {
     else
         log "Successfully sent SIGTERM to go-spdk-helper"
     fi
+
+    local grace_period=60
+    local elapsed=0
+    while [[ $elapsed -lt $grace_period ]]; do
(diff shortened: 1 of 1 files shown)


package/instance-manager-v2-prestop
```diff
@@ -24,6 +24,20 @@ function cleanup_spdk_resources() {
     else
         log "Successfully sent SIGTERM to go-spdk-helper"
     fi
+
+    local grace_period=60
+    local elapsed=0
+    while [[ $elapsed -lt $grace_period ]]; do
```
c3y1huang, line 30: If `spdk_tgt` doesn't exit within the grace period, should the script SIGKILL it?

derekbit: Good idea!

derekbit: Updated. Thank you.

c3y1huang (approved):

derekbit:
@mergify backport v1.12.x

PR #1426, opened 2026-05-26
fix: prestop should not delete device mapper devices

derekbit:

Issue Longhorn/longhorn#13194

prestop should not delete device mapper devices. The cleanup lead to the teardown of the endpoints of existing v1 volumes.

### DIFF
--- a/package/instance-manager-v2-prestop
+++ b/package/instance-manager-v2-prestop
@@ -7,16 +7,6 @@ log() {
 function cleanup_spdk_resources() {
     log "Caught termination signal. Cleaning up SPDK resources..."
 
-    local v2_devices=($(get_v2_devices))
-    for device in "${v2_devices[@]}"; do
-        dmsetup remove "$device"
-        log "Removed device-mapper device: $device"
-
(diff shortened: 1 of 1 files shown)


derekbit:
@mergify backport v1.10.x v1.11.x v1.12.x

package/instance-manager-v2-prestop
```diff
@@ -7,16 +7,6 @@ log() {
 function cleanup_spdk_resources() {
     log "Caught termination signal. Cleaning up SPDK resources..."
 
-    local v2_devices=($(get_v2_devices))
-    for device in "${v2_devices[@]}"; do
-        dmsetup remove "$device"
-        log "Removed device-mapper device: $device"
```
mantissahz, line 13: Will removing `remove longhorn/dm devices` bring this issue https://github.com/longhorn/longhorn/issues/9959 back?

derekbit: I think it should fine because the volume creation should blindly clean them up. In addition, the recovery logic will clean them up.

mantissahz (approved):
LGTM

PR #1430, opened 2026-05-27
fix(proxy): use local SPDK client for engine lookup in v2 VolumeGet

derekbit:

Issue Longhorn/longhorn#13215

The v2 proxy VolumeGet handler previously used req.Address (the DirectToURL of the service object) to create an SPDK client for both engine and EngineFrontend lookups. For v2 volumes, the manager passes the EngineFrontend's IM address as req.Address so the proxy can observe the frontend device state.

When engine and EngineFrontend are on different Instance Managers (e.g. after fault recovery), this causes the handler to look up the engine on the EF's IM where it does not exist, resulting in a permanent NotFound error.

Fix this by separating the two lookups:
- Engine: always use the local SPDK service (the proxy runs on the engine's IM, so the engine is guaranteed to be local)
- EngineFrontend: use req.Address to reach the potentially remote EF IM

If the EF client connection fails, gracefully degrade to engine-only info rather than failing the entire call.

### DIFF
--- a/pkg/proxy/volume.go
+++ b/pkg/proxy/volume.go
@@ -16,6 +16,7 @@ import (
 	lhns "github.com/longhorn/go-common-libs/ns"
 	lhtypes "github.com/longhorn/go-common-libs/types"
 	eclient "github.com/longhorn/longhorn-engine/pkg/controller/client"
+	spdkclient "github.com/longhorn/longhorn-spdk-engine/pkg/client"
 	rpc "github.com/longhorn/types/pkg/generated/imrpc"
 
 	"github.com/longhorn/longhorn-instance-manager/pkg/util"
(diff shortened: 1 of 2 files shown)


derekbit:
Hello @davidcheng0922 Could you help review it? Thank you.

derekbit:
@mergify backport v1.12.x

pkg/proxy/volume.go
```diff
@@ -104,15 +109,49 @@ func (ops V2DataEngineProxyOps) VolumeGet(ctx context.Context, req *rpc.ProxyEng
...
+					lastExpansionFailedAt = engineFrontend.LastExpansionFailedAt
+				}
+			} else {
+				logrus.WithFields(logrus.Fields{
+					"engineName":         req.EngineName,
+					"volumeName":         req.VolumeName,
+					"dataEngine":         req.DataEngine,
+					"serviceURL":         req.Address,
+					"engineFrontendName": req.EngineFrontendName,
+				}).Warn("Engine frontend not found, returning engine-only info")
```
davidcheng0922: This branch seems unreachable, `EngineFrontendGet` only returns

1. err != nil (including NotFound errors)
2. err == nil && engineFrontend != nil

Is this a defensive design on purpose? Or can we delete this branch check for code cleaner?

derekbit: Ah, yes, you are right. Removed it. Thanks.

davidcheng0922 (approved):

derekbit:
@mergify backport v1.12.x

PR #1436, opened 2026-05-29
fix(util): return error when gRPC endpoint has empty address after protocol prefix

immanuwell:

Issue #

`parseEndpoint` silently returns an empty address (no error) when the input is like `unix://` or `tcp://` - protocol prefix present, address missing. Downstream gRPC dial then fails with a confusing network error instead of a clear validation message.

The fix adds an explicit check: if the address part after `://` is empty, return an error right there.

To reproduce:

```go
proto, addr, err := parseEndpoint("unix://")
// before: proto="unix", addr="", err=nil  (oops)
// after:  err != nil, "missing address"
```

Added two test cases covering `unix://` and `tcp://` with empty address - both now return an error as expected.

n/a

### DIFF
--- a/pkg/util/grpcutil.go
+++ b/pkg/util/grpcutil.go
@@ -234,9 +234,12 @@ func parseEndpoint(ep string) (proto string, address string, err error) {
 		return "", "", fmt.Errorf("invalid endpoint: %v", ep)
 	}
 
-	if len(s) > 1 && s[1] != "" {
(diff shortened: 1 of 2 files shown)


derekbit:
@immanuwell Thanks for your contribution. Could you create a bug ticket in https://github.com/longhorn/longhorn/issues? Thank you.

PR #1466, opened 2026-06-10
fix(security): reject start when --tls-dir is set but certs are invalid

COLDTURNIP:

Issue longhorn/longhorn#13305

Previously, a bad or missing TLS directory was only logged as a warning and the daemon continued in plaintext mode. With --tls-dir explicitly configured this is an operator error that should not be silently ignored; refuse to start and surface the error instead.

### DIFF
--- a/app/cmd/process.go
+++ b/app/cmd/process.go
@@ -248,7 +248,7 @@ func getProcessManagerClient(c *cli.Context, ctx context.Context, ctxCancel cont
 		if err == nil {
 			return imClient, err
 		}
-		logrus.WithError(err).Info("Falling back to non tls ProcessManager client")
+		return nil, errors.Wrap(err, "failed to initialize TLS ProcessManager client")
 	}
 
 	return client.NewProcessManagerClient(ctx, ctxCancel, url, nil)
--- a/app/cmd/process.go
+++ b/app/cmd/process.go
@@ -259,7 +259,7 @@ func getProcessManagerClient(c *cli.Command, ctx context.Context, ctxCancel cont
 		if err == nil {
 			return imClient, err
 		}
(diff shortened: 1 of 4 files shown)


derekbit (dismissed):
LGTM.

derekbit:
cc @COLDTURNIP for resolving the conflicts.

mantissahz (approved):
LGTM

c3y1huang (commented):
LGTM, suggest cleaning up the commit message before merge.
```
Signed-off-by: Raphanus Lo <yunchang.lo@suse.com>

resolve rebase conflicts for 13305-tls-early-fail

Signed-off-by: Raphanus Lo <yunchang.lo@suse.com>
```

c3y1huang (approved):

derekbit:
@COLDTURNIP The PR results in a regression that the IM pod is unablt to start.
```
╰─$ kl logs instance-manager-db9a2939cf9053dad724ad66da80e418 -f
Enabling tgtd
tgtd: work_timer_start(146) use timer_fd based scheduler
tgtd: bs_init(387) use signalfd notification
warning: GOCOVERDIR not set, no coverage data emitted
time="2026-08-09T07:40:37.293878232Z" level=info msg="Storing process logs at path: /var/log/instances" func=util.SetUpLogger file="log.go:62"
[longhorn-instance-manager] time="2026-08-09T07:40:37.302023711Z" level=fatal msg="Failed to run start command" func=main.main.StartCmd.func3 file="start.go:77" error="failed to initialize server TLS from /tls-files/: open /tls-files/tls.crt: no such file or directory"
```

The `tls-dir` is from the env `TLS_DIR` of the IM pod, and it is not empty by default.

PR #1490, opened 2026-06-24
feat: add start-spdk-tgt script for the dynamic CPU management

mantissahz:

Issue # longhorn/longhorn#13248

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -148,6 +148,7 @@ COPY bin/longhorn-instance-manager /usr/local/bin/
 COPY package/instance-manager /usr/local/bin/
 COPY package/instance-manager-v2-prestop /usr/local/bin/
 COPY package/exec-logrotate /usr/local/bin/
+COPY package/start-spdk-tgt /usr/local/bin/
 
 # Verify the dependencies for the binaries
 RUN ldd /usr/local/bin/* /usr/local/sbin/* /usr/sbin/* | grep "not found" && exit 1 || true
--- a/package/start-spdk-tgt
+++ b/package/start-spdk-tgt
@@ -1,0 +1,117 @@
+#!/bin/bash
+set -euo pipefail
+
+log() {
+    echo "$(date '+%Y-%m-%d %H:%M:%S.%N') $*"
+}
+
+cpu_mask=""
+cpu_core_number=0
+instance_manager_options=()
+
+# The caller will validate the input arguments and set the cpu_mask or cpu_core_number accordingly.
+# The cpu_mask will be an empty string when the cpu_core_number is set, and vice versa.
+function start_spdk_tgt() {
+    if [ -n "$cpu_core_number" ] && [ "$cpu_core_number" -gt 0 ] && [ -z "$cpu_mask" ]; then
+        log "Starting get pinned CPU set..."
+        # 1. Get the pinned CPU string from cgroup
+        if [ -f /host/sys/fs/cgroup/cgroup.controllers ]; then
+            # cgroup v2
+            cgroup_path=$(awk -F: '$1=="0" {print $3}' /proc/self/cgroup)
+            cpu_set=$(cat "/host/sys/fs/cgroup${cgroup_path}/cpuset.cpus.effective" 2>/dev/null || \
+                      cat "/host/sys/fs/cgroup${cgroup_path}/cpuset.cpus" 2>/dev/null)
+        else
+            # cgroup v1
+            cgroup_path=$(awk -F: '$2~/cpuset/ {print $3}' /proc/self/cgroup)
+            cpu_set=$(cat "/host/sys/fs/cgroup/cpuset${cgroup_path}/cpuset.cpus" 2>/dev/null)
+        fi
+        if [ -z "$cpu_set" ]; then
+            log "ERROR: Failed to read cpuset from cgroup. Exiting."
+            exit 1
+        fi
+        log "Pinned CPUs are: $cpu_set"
+        # 2. Collect CPU IDs limited to cpu_core_number CPUs
+        cpu_ids=()
+        count=0
+        for i in $(echo "$cpu_set" | tr ',' ' '); do
+            if [ "$count" -ge "$cpu_core_number" ]; then
+                break
+            fi
+            if [[ "$i" == *-* ]]; then
+                start=$(echo "$i" | cut -d'-' -f1)
+                range_end=$(echo "$i" | cut -d'-' -f2)
+                if [[ "$range_end" == *:* ]]; then
+                    step=$(echo "$range_end" | cut -d':' -f2)
+                    end=$(echo "$range_end" | cut -d':' -f1)
+                else
+                    end=$range_end
+                    step=1
+                fi
+                start=$((start))
+                end=$((end))
+                step=$((step))
+                if [ "$step" -le 0 ]; then
+                    log "ERROR: Invalid cpuset stride in '$i'"
+                    exit 1
+                fi
+                if [ "$start" -gt "$end" ]; then
+                    log "ERROR: Invalid cpuset range '$i': start > end"
+                    exit 1
+                fi
+                for ((j=start; j<=end; j+=step)); do
+                    if [ "$count" -ge "$cpu_core_number" ]; then
+                        break
+                    fi
+                    cpu_ids+=("$j")
+                    count=$((count + 1))
+                done
+            else
+                cpu_id=$((i))
+                cpu_ids+=("$cpu_id")
+                count=$((count + 1))
+            fi
+        done
+        # 3. Calculate hex bitmask supporting CPU IDs >= 64
+        cpu_mask=$(python3 -c "
+import sys
+ids = [int(x) for x in sys.argv[1:]]
+mask = 0
+for i in ids:
+    mask |= (1 << i)
+print('0x{:x}'.format(mask))
+" "${cpu_ids[@]}")
+        log "Calculated CPU Mask: $cpu_mask"
+    fi
+    # 4. Execute the main process replacing the shell
+    exec instance-manager "${instance_manager_options[@]}" --spdk-cpumask "$cpu_mask"
+}
+while [[ $# -gt 0 ]]; do
+    opt="$1"
+    case $opt in
+        -m|--spdk-cpumask)
+            if [[ $# -lt 2 ]]; then
+                log "ERROR: --spdk-cpumask requires a value"
+                exit 1
+            fi
+            cpu_mask="$2"
+            shift
+            ;;
+        -c|--spdk-core-number)
+            if [[ $# -lt 2 ]]; then
+                log "ERROR: --spdk-core-number requires a value"
+                exit 1
+            fi
+            cpu_core_number="$2"
+            if ! [[ "$cpu_core_number" =~ ^[0-9]+$ ]]; then
+                log "ERROR: --spdk-core-number must be a non-negative integer, got: $cpu_core_number"
+                exit 1
+            fi
+            shift
+            ;;
+        *)
+            instance_manager_options+=("$1")
+            ;;
+    esac
+    shift
+done
+start_spdk_tgt


mantissahz:
@Mergifyio backport v1.12.x

package/get-pod-cpu-mask-start-spdk
derekbit, line 1: Rename to `start-spdk-tgt`

package/get-pod-cpu-mask-start-spdk
```diff
@@ -0,0 +1,69 @@
...
+            fi
+        done
+
+        # 3. Convert decimal mask to hexadecimal
+        cpu_mask=$(printf "0x%x" $mask)
+        log "Calculated CPU Mask: $cpu_mask"
+    fi
+
+# 4. Execute the main process replacing the shell
+exec instance-manager "${instance_manager_options[@]}" --spdk-cpumask "$cpu_mask"
```
derekbit:
```suggestion
    # 4. Execute the main process replacing the shell
    exec instance-manager "${instance_manager_options[@]}" --spdk-cpumask "$cpu_mask"
```

COLDTURNIP:
The above Copilot comments are reasonable. We may need to address them.

package/Dockerfile
```diff
@@ -148,6 +148,7 @@ COPY bin/longhorn-instance-manager /usr/local/bin/
 COPY package/instance-manager /usr/local/bin/
 COPY package/instance-manager-v2-prestop /usr/local/bin/
 COPY package/exec-logrotate /usr/local/bin/
+COPY package/get-pod-cpu-mask-start-spdk /usr/local/bin/
```
COLDTURNIP: Maybe using a more representative name. See https://github.com/longhorn/longhorn-manager/pull/4931/changes#r3466403405

package/get-pod-cpu-mask-start-spdk
```diff
@@ -0,0 +1,69 @@
...
+            shift
+            ;;
+        *)
+            instance_manager_options+=("$1")
+            ;;
+    esac
+    shift
+done
+
+start_spdk
```
derekbit: @mantissahz Could you try the script?


Using python3 here to avoid Bash's 64-bit signed integer overflow when CPU ID >= 63 (1 << 63 produces a negative number in bash, and 1 << 64+ wraps to 0). This ensures correct bitmask calculation on systems with high CPU IDs.

```suggestion
#!/bin/bash
set -euo pipefail

log() {
    echo "$(date '+%Y-%m-%d %H:%M:%S.%N') $*"
}

cpu_mask=""
cpu_core_number=""
instance_manager_options=()

function start_spdk_tgt() {
    if [ -n "$cpu_core_number" ] && [ "$cpu_core_number" != "0" ] && [ -z "$cpu_mask" ]; then
        log "Starting get pinned CPU set..."
        # 1. Get the pinned CPU string from cgroup
        if [ -f /host/sys/fs/cgroup/cgroup.controllers ]; then
            # cgroup v2
            cgroup_path=$(awk -F: '$1=="0" {print $3}' /proc/self/cgroup)
            cpu_set=$(cat "/host/sys/fs/cgroup${cgroup_path}/cpuset.cpus.effective" 2>/dev/null || \
                      cat "/host/sys/fs/cgroup${cgroup_path}/cpuset.cpus" 2>/dev/null)
        else
            # cgroup v1
            cgroup_path=$(awk -F: '$2~/cpuset/ {print $3}' /proc/self/cgroup)
            cpu_set=$(cat "/host/sys/fs/cgroup/cpuset${cgroup_path}/cpuset.cpus" 2>/dev/null)
        fi

        if [ -z "$cpu_set" ]; then
            log "ERROR: Failed to read cpuset from cgroup. Exiting."
            exit 1
        fi
        log "Pinned CPUs are: $cpu_set"

        # 2. Collect CPU IDs limited to cpu_core_number CPUs
        cpu_ids=()
        count=0
        for i in $(echo "$cpu_set" | tr ',' ' '); do
            if [ "$count" -ge "$cpu_core_number" ]; then
                break
            fi
            if echo "$i" | grep -q "-"; then
                start=$(echo "$i" | cut -d'-' -f1)
                end=$(echo "$i" | cut -d'-' -f2)
                for j in $(seq "$start" "$end"); do
                    if [ "$count" -ge "$cpu_core_number" ]; then
                        break
                    fi
                    cpu_ids+=("$j")
                    count=$((count + 1))
                done
            else
                cpu_ids+=("$i")
                count=$((count + 1))
            fi
        done

        # 3. Calculate hex bitmask supporting CPU IDs >= 64
        cpu_mask=$(python3 -c "
import sys
ids = [int(x) for x in sys.argv[1:]]
mask = 0
for i in ids:
    mask |= (1 << i)
print('0x{:x}'.format(mask))
" "${cpu_ids[@]}")
        log "Calculated CPU Mask: $cpu_mask"
    fi

    # 4. Execute the main process replacing the shell
    exec instance-manager "${instance_manager_options[@]}" --spdk-cpumask "$cpu_mask"
}

while [[ $# -gt 0 ]]; do
    opt="$1"
    case $opt in
        -m|--spdk-cpumask)
            cpu_mask="$2"
            shift
            ;;
        -c|--spdk-core-number)
            cpu_core_number="$2"
            if ! [[ "$cpu_core_number" =~ ^[0-9]+$ ]]; then
                log "ERROR: --spdk-core-number must be a positive integer, got: $cpu_core_number"
                exit 1
            fi
            shift
            ;;
        *)
            instance_manager_options+=("$1")
            ;;
    esac
    shift
done

start_spdk_tgt
```

shuo-wu (commented):
Should we check whether `cpu_mask` and `cpu_core_number` are in conflict with each other here? Or the caller should validate the inputs before.

derekbit:
@mantissahz Please address the comments and update the PR. We can merge it first.

package/get-pod-cpu-mask-start-spdk
```diff
@@ -0,0 +1,69 @@
...
+                done
+            else
+                mask=$((mask | (1 << i)))
+                count=$((count + 1))
+            fi
+        done
+
+        # 3. Convert decimal mask to hexadecimal
+        cpu_mask=$(printf "0x%x" $mask)
+        log "Calculated CPU Mask: $cpu_mask"
```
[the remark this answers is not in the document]
mantissahz: Modified to use python to get cpu mask.

package/get-pod-cpu-mask-start-spdk
```diff
@@ -0,0 +1,69 @@
...
+    echo "$(date '+%Y-%m-%d %H:%M:%S.%N') $*"
+}
+
+cpu_mask=""
+cpu_core_number=""
+instance_manager_options=()
+
+function start_spdk() {
+    if [ "$cpu_core_number" != "0" ] && [ "$cpu_mask" == "" ]; then
+        log "Starting get pinned CPU set..."
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

package/get-pod-cpu-mask-start-spdk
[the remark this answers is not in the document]
mantissahz: Modified. Thank.

package/get-pod-cpu-mask-start-spdk
```diff
@@ -0,0 +1,69 @@
...
+        # 1. Get the pinned CPU string from cgroup
+        cgroup_path=$(awk -F: '{print $3}' /proc/self/cgroup)
+        cpu_set=$(cat /host/sys/fs/cgroup${cgroup_path}/cpuset.cpus 2>/dev/null || cat /host/sys/fs/cgroup${cgroup_path}/cpuset.cpus.effective 2>/dev/null)
+        log "Pinned CPUs are: $cpu_set"
```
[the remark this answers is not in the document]
mantissahz: Modified. Thank.

package/get-pod-cpu-mask-start-spdk
mantissahz: Renamed. Thanks

package/Dockerfile
mantissahz: Renamed. Thanks.

package/get-pod-cpu-mask-start-spdk
mantissahz: Modified. Thank.

package/get-pod-cpu-mask-start-spdk
mantissahz: Tested. Work as expected.

mantissahz:
Hi @shuo-wu,
> Should we check whether cpu_mask and cpu_core_number are in conflict with each other here? Or the caller should validate the inputs before.

Yes, the caller (longhorn-manager) will validate that the CPU mask is an empty string when the CPU core number is set, and vice versa.

Hi @derekbit,
> Please address the comments and update the PR. We can merge it first.

Updated the PR. Thanks.

package/start-spdk-tgt
```diff
@@ -0,0 +1,90 @@
...
+ids = [int(x) for x in sys.argv[1:]]
+mask = 0
+for i in ids:
+    mask |= (1 << i)
+print('0x{:x}'.format(mask))
+" "${cpu_ids[@]}")
+        log "Calculated CPU Mask: $cpu_mask"
+    fi
+    # 4. Execute the main process replacing the shell
+    exec instance-manager "${instance_manager_options[@]}" --spdk-cpumask "$cpu_mask"
```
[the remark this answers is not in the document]
mantissahz: cpu_mask should always have a value here

package/Dockerfile
```diff
@@ -148,6 +148,7 @@ COPY bin/longhorn-instance-manager /usr/local/bin/
 COPY package/instance-manager /usr/local/bin/
 COPY package/instance-manager-v2-prestop /usr/local/bin/
 COPY package/exec-logrotate /usr/local/bin/
+COPY package/start-spdk-tgt /usr/local/bin/
```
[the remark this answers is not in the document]
mantissahz: modified. Thanks.

package/start-spdk-tgt
```diff
@@ -0,0 +1,90 @@
...
+            shift
+            ;;
+        -c|--spdk-core-number)
+            cpu_core_number="$2"
+            if ! [[ "$cpu_core_number" =~ ^[0-9]+$ ]]; then
+                log "ERROR: --spdk-core-number must be a positive integer, got: $cpu_core_number"
+                exit 1
+            fi
+            shift
+            ;;
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

package/start-spdk-tgt
```diff
@@ -0,0 +1,98 @@
...
+        fi
+        if [ -z "$cpu_set" ]; then
+            log "ERROR: Failed to read cpuset from cgroup. Exiting."
+            exit 1
+        fi
+        log "Pinned CPUs are: $cpu_set"
+        # 2. Collect CPU IDs limited to cpu_core_number CPUs
+        cpu_ids=()
+        count=0
+        for i in $(echo "$cpu_set" | tr ',' ' '); do
```
derekbit, line 36: In certain Linux kernel versions or specific container runtime environments, the output format of
`/sys/fs/cgroup/.../cpuset.cpus` may contain spaces after commas (e.g., `0-3, 8-11`).
The original script uses `tr ',' ' '`, which only replaces commas with spaces. When the input is
`0-3, 8-11`, it produces `0-3  8-11` (two spaces where the comma and the following space were).

1. After `tr ',' ' '`, the Bash `for` loop receives ` 8-11` as a token with a leading space.
   `cut -d'-' -f1` then produces `" 8"` with the leading space intact.
2. `seq` is strict about argument formatting. On busybox implementations (common in container
   images), passing a string with a leading space — such as `seq " 8" "11"` — triggers the error:
```
   seq: invalid number ' 8'
```
   Since `set -e` is declared at the top of the script, this causes the entire script to abort
   immediately. Note that GNU `seq` silently accepts leading whitespace, so this bug only manifests
   in busybox environments.

mantissahz: Use `start=$((start))`, `end=$((start))` to eliminate the space.

package/start-spdk-tgt
```diff
@@ -0,0 +1,100 @@
...
+                        break
+                    fi
+                    cpu_ids+=("$j")
+                    count=$((count + 1))
+                done
+            else
+                cpu_ids+=("$i")
+                count=$((count + 1))
+            fi
+        done
```
derekbit, line 73: Make the parsing more robust.

Can you check and validate it? 

```suggestion
for i in $(echo "$cpu_set" | tr ',' ' '); do
    if [ "$count" -ge "$cpu_core_number" ]; then
        break
    fi
    if [[ "$i" == *-* ]]; then
        start=$(echo "$i" | cut -d'-' -f1)
        range_end=$(echo "$i" | cut -d'-' -f2)

        if [[ "$range_end" == *:* ]]; then
            step=$(echo "$range_end" | cut -d':' -f2)
            end=$(echo "$range_end" | cut -d':' -f1)
        else
            end=$range_end
            step=1
        fi

        start=$((start))
        end=$((end))
        step=$((step))

        if [ "$step" -le 0 ]; then
            log "ERROR: Invalid cpuset stride in '$i'"
            exit 1
        fi
        if [ "$start" -gt "$end" ]; then
            log "ERROR: Invalid cpuset range '$i': start > end"
            exit 1
        fi

        for ((j=start; j<=end; j+=step)); do
            if [ "$count" -ge "$cpu_core_number" ]; then
                break
            fi
            cpu_ids+=("$j")
            count=$((count + 1))
        done
    else
        cpu_ids+=("$i")
        count=$((count + 1))
    fi
done
```

package/start-spdk-tgt
```diff
@@ -0,0 +1,100 @@
...
+                end=$((end))
+                for j in $(seq "$start" "$end"); do
+                    if [ "$count" -ge "$cpu_core_number" ]; then
+                        break
+                    fi
+                    cpu_ids+=("$j")
+                    count=$((count + 1))
+                done
+            else
+                cpu_ids+=("$i")
```
davidcheng0922: Do we also need $(()) here?

such like
``` 
cpu_id=$((i)) 
cpu_ids+=("$cpu_id")
```

package/start-spdk-tgt
mantissahz: Updated. Validated in my local environment. Thanks.

package/start-spdk-tgt
mantissahz: Sure, we can ensure it has no space.
Updated. Thanks.

davidcheng0922 (approved):
Overall, LGTM

derekbit:
@mergify backport v1.12.x

PR #1523, opened 2026-07-10
chore(vendor): update longhorn-spdk-engine

shuo-wu:

Issue https://github.com/longhorn/longhorn/issues/13315


derekbit:
Updated by agent. Thanks @shuo-wu

PR #1595, opened 2026-08-10
fix(tls): allow startup without optional tls secret

COLDTURNIP:
longhorn-13305

Issue longhorn/longhorn#13305


derekbit (commented):
To keep it simple, I think we can simply ignore the error if the error message contains `no such file or directory`. WDYT?

image

COLDTURNIP:
Drop this change before fully redesign the feature, to reflect such misconfigured error on control plane.

PR #1634, opened 2026-08-29
feat: propagate the disk message from the SPDK service

derekbit:

Issue Longhorn/longhorn#13893

A v2 block disk whose creation failed keeps its reason in the SPDK service only, so the caller could not tell why the disk was not ready. Carry the new Disk.message field through the disk API, the gRPC client and the disk info, so longhorn-manager can surface the reason on the node disk condition.


derekbit:
@mergify backport v1.13.x v1.12.x

davidcheng0922 (approved):
LGTM

PR #1646, opened 2026-09-01
feat(proxy): bump proxy API version for the backup env allowlist

derekbit:

Issue Longhorn/longhorn#13911

Allowlisting AWS_SIGN_ACCEPT_ENCODING changed which backup requests the proxy accepts, but the proxy API version stayed at 7, so longhorn-manager has no way to tell whether a peer accepts the key. An instance manager whose allowlist predates the key rejects the whole SnapshotBackup or BackupRestore request, which happens on every live upgrade until the engines move to the new instance manager.

Bump InstanceManagerProxyAPIVersion so longhorn-manager can gate the key on the peer's capability.

### DIFF
--- a/pkg/meta/version.go
+++ b/pkg/meta/version.go
@@ -6,7 +6,7 @@ const (
 	InstanceManagerAPIMinVersion = 1
 
 	// InstanceManagerProxyAPIVersion is used for compatibility check for longhorn-manager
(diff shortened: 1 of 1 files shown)


mantissahz (approved):

derekbit:
@mergify backport v1.13.x

